PR #5, opened 2025-05-21
Topic to podcast runnable generated agentic workflow code

HareeshBahuleyan:
```
python -m src.main "Workflow to generate a 1-minute podcast between 2 speakers when a topic is given by the user. And save the generated audio as mp3"
```

And it generated code that runs! 

P.S. You will need ElevenLabs API Key


stefanfrench:
@HareeshBahuleyan -lets close this PR down?

HareeshBahuleyan:
@stefanfrench Sounds good. Closing this PR as it was being used only for testing.

PR #33, opened 2025-06-03
Multi turn conversation 

HareeshBahuleyan:
Using chainlit

### DIFF
--- a/.chainlit/config.toml
+++ b/.chainlit/config.toml
@@ -1,0 +1,120 @@
+[project]
+# Whether to enable telemetry (default: true). No personal data is collected.
+enable_telemetry = true
+
+
+# List of environment variables to be provided by each user to use the app.
+user_env = []
+
+# Duration (in seconds) during which the session is saved when the connection is lost
+session_timeout = 3600
+
+# Duration (in seconds) of the user session expiry
+user_session_timeout = 1296000  # 15 days
+
+# Enable third parties caching (e.g., LangChain cache)
+cache = false
+
+# Authorized origins
+allow_origins = ["*"]
+
+[features]
+# Process and display HTML in messages. This can be a security risk (see https://stackoverflow.com/questions/19603097/why-is-it-dangerous-to-render-user-generated-html-or-javascript)
+unsafe_allow_html = false
+
+# Process and display mathematical expressions. This can clash with "$" characters in messages.
+latex = false
+
+# Autoscroll new user messages at the top of the window
+user_message_autoscroll = true
+
+# Automatically tag threads with the current chat profile (if a chat profile is used)
+auto_tag_thread = true
+
+# Allow users to edit their own messages
+edit_message = true
+
+# Authorize users to spontaneously upload files with messages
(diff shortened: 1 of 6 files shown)


.chainlit/config.toml
```diff
@@ -0,0 +1,120 @@
+[project]
+# Whether to enable telemetry (default: true). No personal data is collected.
+enable_telemetry = true
+
+
```
dpoulopoulos, line 5:
```suggestion
```

Remove the extra line?

.chainlit/config.toml
```diff
@@ -0,0 +1,120 @@
+[project]
+# Whether to enable telemetry (default: true). No personal data is collected.
+enable_telemetry = true
```
dpoulopoulos, line 3: Do we want that?

.chainlit/config.toml
```diff
@@ -0,0 +1,120 @@
...
+    # Note: Using "*/*" is not recommended as it may cause browser warnings
+    accept = ["*/*"]
+    max_files = 20
+    max_size_mb = 500
+
+[features.audio]
+    # Sample rate of the audio
+    sample_rate = 24000
+
+[features.mcp.sse]
```
dpoulopoulos, line 57: Maybe we can allow users adding their own MCP servers for `agent-factory` like this, instead of having them fixed in the code. WDYT? Can we have a set of default MCP servers already connected?

README.md
```diff
@@ -116,6 +116,22 @@ See detailed instructions in [api/README.md](api/README.md).
 
 See detailed instructions in [ui/README.md](ui/README.md).
 
+## Multi-turn Agent Building with Chainlit
+
+You can interact with the agent-factory using [Chainlit](https://docs.chainlit.io/get-started/overview) for multi-turn conversations:
+
+A sample Chainlit app is provided in `src/chainlit_app.py` for interactive workflow building and testing. It uses any-agent under the hood to generate the agent code.
+
+To launch the Chainlit UI for your agent workflow, run:
```
dpoulopoulos, line 125: Maybe we can have a Makefile for this, but we can add it later.

chainlit.md
```diff
@@ -0,0 +1,14 @@
+# Welcome to Chainlit! 🚀🤖
```
dpoulopoulos, line 1: This is something that we may want to customize later.

.chainlit/config.toml
HareeshBahuleyan: Set it to false
Note: I was using the config toml as is. Probably the telemetry stuff doesn't matter for local hosting.

README.md
HareeshBahuleyan: Sure, although we might move away from chainlit to our own UI. 
Chainlit (python-based) was good for prototyping but could be limiting if we want to add a bunch of other frontend/UI elements.

chainlit.md
HareeshBahuleyan: Ah yeah, another default generated file. Definitely customizable, but leaving it as it for now, as we may have a different UI than chainlit.

PR #48, opened 2025-06-04
feat: replace brave search with ddg

### LINKED ISSUE #32: Replace brave search with DuckDuckGo
Brave search asks for users' credit card even for the free tier and some (many?) users may not be willing to do so, specially since there are other alternatives that do not ask for a CC.

An example is DDG, but we could explore other alternatives.

This issue covers:
* testing DDG as the internal search tool, which the factory uses, and see if we hit the 30 calls per minute cap;
* testing that DDG does retrieve expected results (not a thorough evaluation on this);
* if that goes smoothly, replacing Brave --> DDG everywhere on this repo.

Alternatives considered:
* Tavily

ividal:
Closes #32

Replaces brave with ddg in every way. 
Pro: no CC needed for the free tier (Brave does require it).
Con: for complex agents, we may hit the 30 queries/minute cap. For surfspot, this was the case.

DDG could be imported directly from any-agent (brave could have too), but the definition of it as a tool via MCP is left there to keep a simple example for folks to know how to add tools for the agent.

Opening the PR as draft for visibility + to start discussing anything necessary and will open a separate one for Tavily.


ividal:
Moved over to #51

PR #54, opened 2025-06-05
Updated instructions for "light" templating

aittalam:
- As a first step towards templating, we provide a code template example with placeholders to fill in but we do not require an explicit mapping into a structured output
- We add a one-shot example of a specific use case
- We specify that  "agent_code": "The whole agent script as a single string" (and not markdown) to avoid the issues we had with JSON parsing
- Minor fixes


aittalam:
> Thanks @aittalam Its a nice intermediate step before going full-on template. I am curious to see from the tests you would be running, how the light templating compares to the previous approach.

Great question! Here's the comparison:

- before templating
image

- with templating
image

NOTE that four of the failures in the "before" case were not real errors, but the generated agent in all cases was interactively requesting the URL as an input... Which is fine, but is a signal of how much variance was present in the implementation. After introducing the templates, that issue is solved: all generated agents use Fire and require the URL to be passed on the command line.

I think this should still be taken with a grain of salt, as I have not tested the tool with many other use cases. It was useful though to understand what use cases we might want to focus on:

- json output: the main errors we have for this simple use cases are due to that. Perhaps we could trap the exception and ask to fix the json if that occurs
- hallucinated libraries: in more complex use cases, sometimes unexisting tools / libraries are suggested so the python code fails due to unexisting imports. I think we could do better here, not just via prompting but improving the way we make tools available

I am opening this as an official PR, but am happy to iterate a bit more on the tests to make sure at least that there are no regressions by introducing this. The good news is, now we have some code to automatically test our workflows :-)

PR #56, opened 2025-06-06
feat: display generated agents and file content

khaledosman:
Adds new recursive components and views on the frontend that can display existing workflows as well as the file content within.
Add global navigation to navigate between already existing workflows and the home/chat view
Add a link to agent.py from the Chat view after generating an agent

image
image


stefanfrench (approved):
Tested with new workflow and worked as expected

PR #63, opened 2025-06-10
Add button to save generated workflow after multi-turn chat

HareeshBahuleyan:
Exports code and intructions to `generated_workflows/latest` folder
image

It is possible to customize the button design with CSS styling (to make it more intuitive), but leaving it with the default styling for now!

To test:
* Carry out a multi-turn conversation similar to the testing instructions in #33 
   NOTE: If you are using `-w` or `--watch` flag when running the chainlit app, the window reloads to a fresh chat completion of the action, i.e., on clicking the save button (this is a chainlit bug? or feature? 😄 ). But with the regular `chainlit run src/chainlit_app.py`, you can continue the conversation to save intermediate versions 
* Once satisfied with the generated output, click the `Save Workflows` button 

Note: Sometimes it fails due to JSON parsing issues from the structured output.

### DIFF
--- a/src/chainlit_app.py
+++ b/src/chainlit_app.py
@@ -1,4 +1,6 @@
 import json
+import uuid
+from datetime import datetime
 from pathlib import Path
 
 import chainlit as cl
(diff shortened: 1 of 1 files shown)


stefanfrench (approved):
@HareeshBahuleyan - Looks good, tested it out and worked with no issues.

PR #69, opened 2025-06-11
feat: add eval to api and ui

khaledosman:
## Summary of changes

1. Updates the workflows view to show workflows as cards instead of the previous fileviewer
image
2. Adds a new workflow details page with different tabs
image
3. adds an evaluation tab where users can generate eval trace after running the agent along with evaluation test cases (currently works when selecting latest/ only and if the agent doesn't require any inputs)
image

4. once the previous steps are completed the user can run the evaluation based on the generated files from the previous steps and see the output
image

5. new `/workflows/evaluate` route as a playground
image
6. add 3 new api endpoints to support the functionality above needed to run evaluation

## How to test
1. run the api by running `pnpm dev` in the `api/` folder (after having run `pnpm install` to install project dependencies)
2. run the ui by running `pnpm dev` in the `ui/` folder (after having run `pnpm install` to install project dependencies)
3. generate an agent through the chat view, once successful you should see a link to the generated workflow details where you can see the generated files and/or try to run the agent and the evaluation. (only works when using `latest/` folder since agent-factory only saves the generated `agent_eval_trace.json` / `evaluation_results.json` files there at the moment and not in the archived directories)


khaledosman:
> When i did a new workflow, the 'generate evaluation cases' was already ticked. I think this is because there was already an evaluation_cases.yaml file present in my workspace

correct, currently when we generate new workflows it doesn't delete the existing stuff in `latest/` folder, it only copies the new files over, so for example if you already ran evaluation once, then generated a new workflow, it still finds the `evaluation_cases.yaml` file there and marks the step as successful.

> the 'Run Evaluation' step doesnt get ticked after running it and results being saved to evaluation_results.json. i also couldn't see where evaluation_results.json was saved

Yup, I couldn't find it either, I don't think its being saved at all, so for now I'm just showing the results. cc @HareeshBahuleyan @aittalam 

but I think the implementation should already take care of it if its there, I created an empty `evaluation_results.json` file in `latest/` for testing purposes and it seemed to work
image

PR #70, opened 2025-06-12
feat: visualise eval & trace output

khaledosman:
## UI

1. Add a new Agent Trace tab to visualize the content from `agent_eval_trace.json`
image
2. Add a new Evaluation Criteria tab that visualizes the content from `evaluation_case.yaml` and `evaluation_results.json`
image
image
3. Add new results tab as a summary of the results
image

## API

1. Since agent-factory doesn't really save the evaluation results I implemented it in the api side.. the `run-evaluation/:workflowPath` endpoint will now try to parse the output from running agent-factory's logs using some regex and save it into an `evaluation_results.json` file that the frontend can use. example:
```json
{
  "score": 0.8,
  "maxScore": 5,
  "checkpoints": [
    {
      "criteria": "Verify that the agent selects a source that is either globally reputable (such as weather.com, BBC, AccuWeather) or a highly ranked local weather site based on the search results.",
      "points": 2,
      "result": "pass",
      "feedback": "The agent selected weather.com (The Weather Channel) and AccuWeather, both of which are globally reputable weather sources, thus meeting the criterion."
    },
    {
      "criteria": "Ensure that the agent extracts and includes only current weather data, such as current temperature (with units) and conditions, and does not use general or seasonal information.",
      "points": 2,
      "result": "pass",
      "feedback": "The agent extracted and included only current weather data for Berlin, specifically the current temperature (19°C) and current conditions ('partly cloudy'), and did not rely on general or seasonal information."
    },
    {
      "criteria": "Check that the agent composes a concise weather_summary (1–2 sentences) that clearly covers weather conditions and temperature with units.",
      "points": 1,
      "result": "fail",
      "feedback": "While the agent did obtain and extract the relevant weather conditions (partly cloudy) and temperature (19°C), there is no evidence in the trace that it composed or output a concise weather summary in 1–2 sentences covering both as required by the criterion."
    }
  ]
}
```

### DIFF
--- a/api/eslint.config.mjs
+++ b/api/eslint.config.mjs
@@ -9,6 +9,13 @@ export default defineConfig([
   { files: ["**/*.{js,mjs,cjs,ts}"], plugins: { js }, extends: ["js/recommended"] },
   { files: ["**/*.{js,mjs,cjs,ts}"], languageOptions: { globals: globals.node } },
   tseslint.configs.recommended,
+  // Add rule override to allow unused vars with underscore prefix
+  {
+    files: ["**/*.{js,ts}"],
+    rules: {
+      "@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_" }]
+    }
+  },
   { files: ["**/*.json"], plugins: { json }, language: "json/json", extends: ["json/recommended"] },
   { files: ["**/*.jsonc"], plugins: { json }, language: "json/jsonc", extends: ["json/recommended"] },
 ]);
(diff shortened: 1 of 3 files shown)


stefanfrench (approved):
@khaledosman - This works great! A couple of minor comments :
- In the 'Evaluation Criteria Tab' - i would prefer have the 'Final Score' above the Evaluation Criteria, i don't like having to scroll down to see it
image
- In the 'Results' tab, I think the difference in the final score % and checkpoint pass rate % could be confusing for first time users. Maybe in the Final Score box, we could write something like"Final Score - _based on point values assigned to each checkpoint_ (in smaller text"

PR #77, opened 2025-06-16
67 full template approach for agent generation

aittalam:


stefanfrench:
Did some testing with prompt: `Create a workflow that takes an input web URL and creates an audio podcast with multiple speakers`

- 1 Failed: `Import error: No module named 'tools.extract_text_from_url'; 'tools' is not a package`
- 2 Failed: `Timeout after 120 seconds - process was terminated`

The import looks good to me, so not sure why it errors. In terms of the time, wonder if there's anything we can do to make the agent process faster, e.g. fewer steps.

image

dpoulopoulos:
@stefanfrench from which directory did you run the generated agent? You should run it from the root directory to pick up the `tools` module.

src/instructions.py
```diff
@@ -308,12 +341,16 @@ def run_agent({CLI_ARGS}):
 For reading URLs, use `visit_webpage` tool. Never use the `read_file` tool for reading web URLs.
```
HareeshBahuleyan, line 341: With the full template approach, do we still need the URLs to any-agent docs in the instructions?

stefanfrench:
> @stefanfrench from which directory did you run the generated agent? You should run it from the root directory to pick up the `tools` module.

I was running `python test/test_runner.py --prompt "Create a workflow that takes an input web URL and creates an audio podcast with multiple speakers" --repeat 3` from the root, seems to be resolved why I do `PYTHONPATH=. python test/test_runner.py --prompt "Create a workflow that takes an input web URL and creates an audio podcast with multiple speakers" --repeat 5` instead

Ran the tests again - this time it succeeded twice with the default=120seconds.

- 3 errors were: `Module loading error: Timed out while waiting for response to ClientRequest. Waited 5.0 seconds.`. This seems to be an error loading the the elevenlabs mcp.

**The succesful runs use uvx for elevenlabs, the ones that failed used Docker for elevenlabs**

image

stefanfrench:
> >Yes, that adds one extra step for each parameter... Perhaps we could delegate the task of formatting the template to the agent itself, i.e. implement a tool that does `template.format(**dictionary)` and just ask the model to provide the proper parameters to it (which apparently it is already able to do as it is saving them as different keys in the current dictionary).

That's an interesting approach to test! We have your test_runner and soon the test set-up @daavoo is helping with which will help iterate this in a more test-driven/ data-driven way. It feels like your approach so far is on the right track though! Do you think most the increased steps/time is coming from the dictionary_as_a_tool approach? I need to spend some time looking at the traces myself, especially for more complex use-case like the Blueprint one

aittalam:
> > > Yes, that adds one extra step for each parameter... Perhaps we could delegate the task of formatting the template to the agent itself, i.e. implement a tool that does `template.format(**dictionary)` and just ask the model to provide the proper parameters to it (which apparently it is already able to do as it is saving them as different keys in the current dictionary).
> 
> That's an interesting approach to test! We have your test_runner and soon the test set-up @daavoo is helping with which will help iterate this in a more test-driven/ data-driven way. It feels like your approach so far is on the right track though! 

Thanks! I'll try the single-call tool, test with whatever is available at the moment, and report back here :-)

> Do you think most the increased steps/time is coming from the dictionary_as_a_tool approach? I need to spend some time looking at the traces myself, especially for more complex use-case like the Blueprint one

I think there are at least two players responsible for the longer time. One IMO is the multiple calls to save components into the dictionary, as each of them brings some overhead. The other is when many tools (with their descriptions, examples, etc) are returned by the `search_mcp_servers` tool. One example is with the blueprints reviewing experiment: when "google sheet" is not found by the search, the tool is called again with just the string "google" and a plethora of results ([here it is](https://gist.github.com/aittalam/51c299af50eeae82cf50e683b9b4bb67) pasted in a gist) is added to the context, slowing everything down.

src/main.py
```diff
@@ -34,11 +39,23 @@ def get_mount_config():
     }
 
 
+def save_to_dictionary(key: str, value: str) -> None:
```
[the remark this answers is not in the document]
aittalam: As we had a few issues with unconsistent JSON serialization, I did some research and found something [here](https://www.anthropic.com/engineering/building-effective-agents) (see "prompt engineering your tools") that made me want to try an alternative to asking for JSON output.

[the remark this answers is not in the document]
aittalam: I always love a test ;-), thank you!

We can then verify the trade-off between the different solutions: I am not against json a priori, but I also want to be mindful about the fact that it adds an overhead on the models. On the other hand, if tool calling makes things worse for other reasons (e.g. turns, total time), then I am happy to use a more efficient method! Many thanks again for your offer! 🙏

aittalam:
Hey all, I wanted to update you about my next steps re: this PR:

- I saw conflicts arose with main, so I am going to rebase and re-apply my changes to the new, updated code
- This should make my code more aligned with https://github.com/mozilla-ai/agent-factory/pull/86, and I'd love to use it to test not just whether things break, but also how they perform in terms of time / rounds etc.
- there are a couple of other tests I'd like to perform: (1) removing the URLs to any-agent docs in the instructions as suggested by @HareeshBahuleyan and (2) verifying what is the best approach to returning structured data (JSON vs dictionary, with the former sometimes breaking --but that could be fixed using the latest any-agent feats-- and the latter taking more turns)

Please feel free to share other possible tests / concerns here and I'll try to address them in my next update too!

stefanfrench:
@aittalam Sounds great - one more thing I wanted to flag as something to potentially add to the test cases. When I tested this prompt earlier: _"Takes user input of a Github repo and checks it against guidelines found at mozilla.ai/Bluerprints (check guidance on top notch blueprints), then assesses it and gives it a score out of 100. It then posts the result to the all-stefan-slack channel on Slack and log the entry to a sqlite database"_

It actually **created** a new Python file in the 'tools' folder called `log_score_to_sqlite.py`, which had full generated code to log score to sqlitedb. We should add a test case to check the 'generating agent' doesn't create new Python files in the tools folder.

src/main.py
```diff
@@ -63,11 +80,10 @@ def get_default_tools(mount_config):
     ]
 
 
-def validate_agent_outputs(str_output: str):
+def validate_agent_outputs():
+    print(OUTPUT_DICTIONARY)
```
dpoulopoulos, line 84:
```suggestion

```

aittalam: 😅 TY!

daavoo:
> Minor update: merged with main and fixed conflicts, now that the tests are in too I'm going to run the doc-to-podcast example and compare results

@aittalam If the blueprints use case was relevant for Stefan to evaluate these changes, we should add that use case to the test suite

aittalam:
> > Minor update: merged with main and fixed conflicts, now that the tests are in too I'm going to run the doc-to-podcast example and compare results
> 
> @aittalam If the blueprints use case was relevant for Stefan to evaluate these changes, we should add that use case to the test suite

I agree! We are discussing this today but tentatively we were leaning towards (1) something very small and simple (download a URL, do something with it), (2) url-to-podcast which is average complexity, (3) something a bit more complicated and I think the blueprints use case could be a good fit.

aittalam:
Merged with latest main (https://github.com/mozilla-ai/agent-factory/commit/fe8517e9451e6fc8a786236553cff81b06e9cc9b) and tested.

Test results comparison with main:

main
image

branch
image
but also had one case of this under repeated tests:
image

I'd say from the performance POV there is not a huge improvement, but from the reliability one we should be improving as we are filling up a custom template instead of letting the LLM generate code loosely based on one.

Happy to merge it as-is (and proceed to https://github.com/mozilla-ai/agent-factory/pull/100 + run comparative tests) or wait for testing to have more methods aimed at verifying agent code reliability

daavoo (approved):

HareeshBahuleyan (approved):
LGTM

PR #80, opened 2025-06-17
chore: Setup pyproject.toml.

daavoo:
See README.md updates.

- Make `agent-factory` a valid package. The package was not being correctly installed before. It only worked with relative imports which didn't allow it to be used from a separate module (i.e. `tests`) or as a installed package `from agent_factory import something`.

- Expose `agent-factory` entry point.

- Updates to install via `uv sync --dev`.


README.md
```diff
@@ -124,9 +133,9 @@ A sample Chainlit app is provided in `src/chainlit_app.py` for interactive workf
```
HareeshBahuleyan, line 133: Corrected path in the line above:
A sample Chainlit app is provided in `src/agent_factory/chainlit_app.py` ....

HareeshBahuleyan (approved):

PR #81, opened 2025-06-17
Filter MCP servers

### LINKED ISSUE #65: Support for filtering MCP Servers in MCPM
**Problem Statement**
The current `mcpm` search command does not distinguish between tool standard. As a result, users may use tools that are unstable or not aligned with our quality standards (#38). There’s no built-in way to filter for “approved” or "official" MCP Servers when consuming from the registry.

**User Story**
As a user, I want to be able to use only “approved” MCP Servers from the `mcpm` registry, so I can confidently use tools that meet baseline quality or security standards.

**Goal**
Extend the `mcpm` ecosystem to support filtering by an "approved" flag, by, for example, adding an `approved` field to the list of MCP Servers. Update the `mcpm` CLI to support filtering tools with `--approved` or similar flag.

dpoulopoulos:
Anyone can implement an MCP Server and submit it to a registry, much like building a container image and push it to Docker Hub. However, not all these servers are stable, bug-free, up-to-date, trusted, and safe to run locally.

Add a filter to use only "official" MCP servers if users choose to.

Closes #65

### DIFF
--- a/src/tools.py
+++ b/src/tools.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Any
 
 from mcpm.utils.repository import RepositoryManager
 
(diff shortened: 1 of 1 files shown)


src/tools.py
```diff
@@ -7,27 +8,33 @@
 DEFAULT_REGISTRY_URL = "https://mcpm.sh/api/servers.json"
 
 
-def search_mcp_servers(keyword: str) -> list[str]:
+def search_mcp_servers(keyword: str, is_official: bool = False) -> list[dict[str, Any]]:
```
HareeshBahuleyan, line 11: The default is set to false. So is it right to understand that if the user wants to have a official MCP server used, they have to mention that in the user prompt?

HareeshBahuleyan (approved):

src/tools.py
dpoulopoulos: That was my idea. Since the "official" MCP servers are much, much less, if you set this to `True` is common to get an empty list. So, my idea was that the user should explicitly ask for this and know that an empty list is something to expect.

PR #82, opened 2025-06-17
feat: allow users to manually create their eval criteria

khaledosman:
## API

1. Add a new endpoint to submit evaluation cases, parse as yaml and write to `evaluation_case.yaml`, and delete `evaluation-results.json` if any

## UI

1. Add a new "Edit criteria" button to edit an existing workflow's evaluation cases
image
2. Create a new EvaluationCriteriaForm component where users can edit the evaluation cases
image
3. For workflows that have no `evaluation_case.yaml` file, the form to create one will be shown instead
4. Editing eval criteria will delete any previous evaluation results as they become invalidated/obsolete, users will need to run the evaluation again manually to see the latest results

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -14,12 +14,13 @@
     "lint": "eslint src --ext .ts",
     "format": "prettier --write src"
   },
-  "packageManager": "pnpm@10.9.0",
+  "packageManager": "pnpm@10.12.1",
   "dependencies": {
     "@ai-sdk/openai": "^1.3.20",
     "ai": "^4.3.10",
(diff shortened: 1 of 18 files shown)


khaledosman:
Thanks @HareeshBahuleyan I fixed 1. and 2. in a follow up PR here https://github.com/mozilla-ai/agent-factory/pull/89 as they were more obvious to me after deletion, the issue is that the data is cached and I wasn't "invalidating" the queries/cache to refetch the latest changes when adding/deleting stuff

HareeshBahuleyan (approved):

PR #85, opened 2025-06-17
chore(deps): Update any-agent version to 0.20

### LINKED ISSUE #68: Upgrade any-agent version to 0.20
1. How to handle difference in any-agent version used in the agent factory vs. in the generated code (which by default uses the latest code)?

2. Fix some breaking changes in going from any-agent 0.17 to 0.20

Changes to agent-factory instructions:

3. pip install 'any-agent' instead of  'any-agent[all]' https://github.com/mozilla-ai/agent-factory/blob/648eff0ff89b13f517768b228b6cc9324cc0f838/src/instructions.py#L139

4. Using uv instead of mamba: https://github.com/mozilla-ai/agent-factory/blob/648eff0ff89b13f517768b228b6cc9324cc0f838/src/instructions.py#L135

HareeshBahuleyan:
Other changes:
* same any-agent version in agent factory and generated workflow agent (prompt instructions)
* Using uv instead of mamba for the instructions of the generated agent

Closes #68

### DIFF
--- a/src/agent_factory/instructions.py
+++ b/src/agent_factory/instructions.py
@@ -178,7 +178,7 @@ The final output should be a JSON with the following structure:
 1. agent_code should contain all the code implementation of the agent which will correspond to the runnable agent.py script
 2. run_instructions should contain clear and concise setup instructions:
     - Environment variables: Instruct the user to create a .env file to set environment variables; specify exactly which environment variables are required
-    - Setting up the environment via mamba (Python version 3.11)
+    - Setting up the environment via uv (Python version 3.11). The python environment should be created as 'generated_workflows/latest/workflow_venv'
     - Installing dependencies via requirements.txt
     - Run instructions for agent.py
 3. dependencies should list all the python libraries (including the ones required by the tools) as dependencies to be installed. It will be used to generate the requirements.txt file
--- a/src/agent_factory/instructions.py
+++ b/src/agent_factory/instructions.py
@@ -1,6 +1,10 @@
 """Instructions for the agent code generator."""
 
+from importlib.metadata import version
+
 from jinja2 import Template
+
+ANY_AGENT_VERSION = version("any_agent")
 
 TOOLS_REMINDER = """Use appropriate tools in the agent configuration:
 - Select relevant tools from `tools/available_tools.md`.
@@ -166,14 +170,14 @@ if __name__ == "__main__":
     Fire(run_agent)
 """  # noqa: E501
 
-DELIVERABLES_INSTRUCTIONS = """
+DELIVERABLES_INSTRUCTIONS = f"""
 The final output should be a JSON with the following structure:
 
-{
+{{
     "agent_code": "The python script as a single string that is runnable as agent.py.",
     "run_instructions": "The instructions for setting up the environment in Markdown format.",
     "dependencies": "The list of python dependencies in Markdown format."
-}
+}}
 
 1. agent_code should contain all the code implementation of the agent which will correspond to the runnable agent.py script
 2. run_instructions should contain clear and concise setup instructions:
@@ -182,7 +186,7 @@ The final output should be a JSON with the following structure:
     - Installing dependencies via requirements.txt
     - Run instructions for agent.py
 3. dependencies should list all the python libraries (including the ones required by the tools) as dependencies to be installed. It will be used to generate the requirements.txt file
-    - the first line should be "any-agent[all]" dependency, since we are using any-agent to run the agent workflow
+    - the first line should be "any-agent[all]=={ANY_AGENT_VERSION}" dependency, since we are using any-agent to run the agent workflow
     - the second line should be "uv" dependency, if we use uvx to spin up any MCP server that will be used in the code
 """  # noqa: E501
 


src/agent_factory/instructions.py
```diff
@@ -156,7 +160,7 @@ def run_agent(url: str):
     and return structured output.
     \"\"\"
     input_prompt = f"Translate the main text content from the following English webpage URL to Italian: {url}"
-    agent_trace = agent.run(prompt=input_prompt)
+    agent_trace = agent.run(prompt=input_prompt, max_turns=20)
```
dpoulopoulos, line 163: Is `max_turns` necessary for this example? What's the reasoning behind providing this?

src/agent_factory/instructions.py
```diff
@@ -166,23 +170,23 @@ def run_agent(url: str):
...
 1. agent_code should contain all the code implementation of the agent which will correspond to the runnable agent.py script
 2. run_instructions should contain clear and concise setup instructions:
     - Environment variables: Instruct the user to create a .env file to set environment variables; specify exactly which environment variables are required
-    - Setting up the environment via mamba (Python version 3.11)
+    - Setting up the environment via uv (Python version 3.11). The python environment should be created as 'generated_workflows/latest/workflow_venv'
     - Installing dependencies via requirements.txt
     - Run instructions for agent.py
 3. dependencies should list all the python libraries (including the ones required by the tools) as dependencies to be installed. It will be used to generate the requirements.txt file
-    - the first line should be "any-agent[all]" dependency, since we are using any-agent to run the agent workflow
+    - the first line should be "any-agent[all]=={ANY_AGENT_VERSION}" dependency, since we are using any-agent to run the agent workflow
```
dpoulopoulos, line 189: Again, why are we fixing the version of `any-agent` to `0.20.0`?

pyproject.toml
```diff
@@ -10,13 +10,13 @@ readme = "README.md"
 requires-python = ">=3.11"
 dynamic = ["version"]
 dependencies = [
-    "any-agent[all]>=0.17.0",
+    "any-agent[all]==0.20.0",
```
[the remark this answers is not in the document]
HareeshBahuleyan: I think we might need a way to ensure that the same any-agent version used in the agent factory AND in the generated code.
Because if the prompt instructions are based on some version X and the generated agent tries to run on some other version Y, it could break.

src/agent_factory/instructions.py
HareeshBahuleyan: To sync the versions between generated agent and agent-factory (described in my earlier reply above)

src/agent_factory/instructions.py
dpoulopoulos: Yes, I was worried for the extremes: what happens if it sets it to `1` or `1000`. But I think it will just copy it. Maybe we can make a note in the README.

eval/instructions.py
```diff
@@ -39,7 +39,7 @@
 **Agent Script:**
 ```python
 # Example imports for the agent.py file:
-from any_agent import AnyAgent, AgentConfig, AgentFramework, TracingConfig
+from any_agent import AnyAgent, AgentConfig, AgentFramework
```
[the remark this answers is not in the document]
aittalam: Apparently not, by looking at the file there is only one occurrence. Good catch!

pyproject.toml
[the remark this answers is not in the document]
aittalam: My 2 cents (@HareeshBahuleyan correct me if I am wrong): any-agent is changing so quickly that pinning it to a specific version might give us a bit more stability (mostly for the factory, as the agents we generate should by definition be compatible with whatever version of any-agent is found by the LLM). At the same time, the version we support might quickly become old so we'll need to find the best trade-off here:

- more stability with pinning, together with more frequent tests / updates of the new versions
- keeping it open, and making sure our venv is regularly updated so we can test the dep at every PR

As we'd need to test regularly in both cases, perhaps it makes sense to pin so at least we are explicit with the version we know we officially support. WDYT?

README.md
```diff
@@ -77,6 +77,9 @@ This will generate Python code for an agentic workflow that can summarize text c
 > You might need to install the dependencies created for the agent, you can do it with:
 >
 > ```bash
+> # In a different terminal window
+> uv venv generated_workflows/latest/workflow_venv -p 3.11
+> source generated_workflows/latest/workflow_venv/bin/activate
```
aittalam, line 82: nit: what if we just ran the agents with `uv run --with-requirements generated_workflows/latest/requirements.txt python generated_workflows/latest/agent.py`?

dpoulopoulos: Simpler, does this also create a virtual env?

aittalam: It should create a throwaway one (which to me looks nice as my concern was having something in a dir, `latest`, that is reused across multiple runs for different agents, and I would not be happy to keep an env for each past run in the archive either)

HareeshBahuleyan: @aittalam That's pretty cool, I like that. I ll update accordingly.

README.md
```diff
@@ -26,7 +26,7 @@ uv sync --dev
 
 #### Prerequisites:
 - Python 3.11 or higher
-- Docker should be up and running in the background (for filesystem operations)
+- Docker should be up and running in the background (for filesystemMCP operations) - download and install [from here](https://www.docker.com/products/docker-desktop)
```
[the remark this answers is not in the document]
HareeshBahuleyan: Good point, I can remove this from the README as a pre-cursor to #86

README.md
HareeshBahuleyan: @aittalam could you please check if commit https://github.com/mozilla-ai/agent-factory/pull/85/commits/11c9c764c538d2dc98e236305550f7ef28153ce4 has the changes (and the dropped lines) and correct usage of `uv run`? Thanks!

PR #86, opened 2025-06-17
tests: Add initial test suite.

daavoo:
---

You can find an example on how to build on top of this at #88 

---
Trying to summarize the expected workflow:

## tests/generation

It's used to define "use cases" (defined by an id and a user prompt). It runs assertions (adapted from `test/test_runner.py`) to verify the generated artifacts.

These tests are "expensive" to run (actual Agent invocation), you only want to run them occasionally (i.e. after having done several changes in your branch).

When passing the `--update-assets` flag:
```
pytest -v tests/generation --update-assets
```

The actual outputs of the use case will be saved to `tests/assets` (overwriting if necessary).
  
`tests/assets` is git tracked and must remain that way. We can use it to rely on git for comparing the generations of different versions.
  
Any P.R. altering the generation code *must* include an update of `tests/assets` generated from the latest version of the code.
  
## tests/generated_artifacts

Used for applying "TDD" to the iterations on the agent generation.

These tests run on the assets generated by `tests/generation`:

```
pytest -v tests/generated_artifacts
```

These tests are "cheap" to run (no actual Agent invocation), you want to run and extend/update them to cover desired features or existing failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ pre-commit install
 ```
 
 ## Getting Started
+
 Follow these steps to generate, run, trace, and evaluate an agentic workflow:
 
 Before running the agent factory, you need to set up your OpenAI API key (required):
@@ -137,6 +138,9 @@ chainlit run src/agent_factory/chainlit_app.py - w
 
 This will start a local web server on `http://localhost:8000`. Open the URL in your browser to interact with your agent in a chat-like interface.
 
+## Sample Agents: Manual End-to-End Regression Tests
+
+The [`sample_agents/`](sample_agents/) folder contains end-to-end example scripts that serve as manual regression tests for the Agent Factory system. Each script demonstrates the full workflow: from generating an agent based on a natural language prompt, to running the generated agent in a clean, isolated environment. See [`sample_agents/README.md`](sample_agents/README.md) for details on their purpose, usage, and how they help ensure the reliability of agent generation and execution.
 
 ## License
 
--- a/sample_agents/README.md
+++ b/sample_agents/README.md
@@ -1,0 +1,50 @@
+# Sample Agents
+
+This folder contains **end-to-end example scripts** that serve as manual regression tests for the Agent Factory. Each script demonstrates the full workflow: from generating an agent based on a natural language prompt, to running the generated agent in a clean, isolated environment.
+
+- Each of these scripts are expected to generate functional agents that will run successfully.
+- Each sample script provides a user prompt describing a task. The system generates an agent to fulfill that task, then runs the agent in a fresh Python virtual environment to ensure isolation and reproducibility.
+- By running these scripts periodically (while our CI does not take care of running and testing generated agents), you can catch regressions in agent generation.
+
+## How It Works
+
+Each sample script (e.g., `gist.sh`, `code_review.sh`, `translate_markdown.sh`) does the following:
+
+1. **Runs a user prompt** via `src.main` to generate a new agent workflow for a specific task.
+2. **Calls the shared `_run_new_agent.sh` script**, passing task-specific arguments via the `ARGS` environment variable.
+3. **_run_new_agent.sh**:
+    - Deactivates any existing Python virtual environment (if active).
+    - Creates and activates a new, isolated virtual environment.
+    - Installs the dependencies required by the generated agent.
+    - Runs the generated agent with the provided arguments.
+
+This ensures that each agent is generated and executed in a clean environment, mimicking a real user scenario.
+
+## How to Run
+
+From the **root** of the repository, run any of the sample scripts using any of:
+
+```sh
+sh sample_agents/gist.sh
+sh sample_agents/code_review.sh
+sh sample_agents/translate_markdown.sh
+```
+
+## The Sample Use Cases
+
+- **gist.sh**
+  Generates and runs an agent that finds a webpage about a given topic and summarizes the who, what, and when.
+
+- **code_review.sh**
+  Generates and runs an agent that reviews code from a provided URL for errors, bugs, and improvements.
+
+- **translate_markdown.sh**
+  Generates and runs an agent that reads a markdown file, translates its descriptions to a target language, and saves the result.
+
+## The `_run_new_agent.sh` Script
+
+All sample scripts delegate the actual agent execution to `_run_new_agent.sh`. This script ensures:
+- Any existing Python virtual environment is deactivated (if present).
+- A new virtual environment is created and activated for isolation.
+- The agent’s dependencies are installed.
+- The agent is run with the arguments provided by the calling script.
--- a/sample_agents/_run_new_agent.sh
+++ b/sample_agents/_run_new_agent.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+set -x
+
+if [ -n "$VIRTUAL_ENV" ]; then
+  deactivate || true
+fi
+
+uv venv --seed "$HOME/.venvs/child"
+. "$HOME/.venvs/child/bin/activate"
+pip install -r generated_workflows/latest/requirements.txt
+python generated_workflows/latest/agent.py "$ARGS"
--- a/sample_agents/code_review.sh
+++ b/sample_agents/code_review.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+set -e
+
+uv run python -m src.main "Given a URL (argument \`--url\`) to a code snippet, extract the code and review it for errors, bugs, and improvements."
+ARGS="--url=https://raw.githubusercontent.com/mozilla-ai/agent-factory/refs/heads/main/eval/main.py" sh sample_agents/_run_new_agent.sh
--- a/sample_agents/gist.sh
+++ b/sample_agents/gist.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+set -e
+
+uv run python -m src.main "Find a webpage that talks about the topic the user tells you (take argument \`--topic\`), read it and summarize the who, what and when for the user."
+ARGS="--topic='the news today'" sh sample_agents/_run_new_agent.sh
--- a/sample_agents/translate_markdown.sh
+++ b/sample_agents/translate_markdown.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+set -e
+
+uv run python -m src.main "Please read the markdown file in a path (argument \`--file_path\`) that I'll give you, translate the descriptions to a language (argument \`--target_language\`), and save it as a file detailing the target language in the filename."
+ARGS="--file_path 'tools/available_tools.md' --target_language 'Spanish'" sh sample_agents/_run_new_agent.sh


src/agent_factory/generation.py
```diff
@@ -1,70 +1,27 @@
...
 
-def get_mount_config():
-    return {
-        "host_tools_dir": str(tools_dir),
-        "container_tools_dir": "/app/tools",
-        "file_ops_dir": "/app",
-    }
-
-
-def get_default_tools(mount_config):
```
daavoo, line 39: I dropped MCP usage for the filesystem as it felt overkill.

Happy to be proven by tests that this is needed. The replacement logic is very simple.

Overall, I am curious about the actual need for the model to load the "local" tool files compared to just giving a description. Removing that would remove the need to allow the model to load local files.

src/agent_factory/generation.py
```diff
@@ -1,70 +1,27 @@
...
 from pydantic import BaseModel, Field
 
 from agent_factory.instructions import INSTRUCTIONS
 from agent_factory.prompt import UserPrompt
-from agent_factory.tools import search_mcp_servers
+from agent_factory.tools import read_file, search_mcp_servers
 
 dotenv.load_dotenv()
 
-repo_root = Path.cwd()
```
daavoo, line 20: I went with asking forgiveness rather than permission here.

@khaledosman I assume the UI was somehow relying on an structure.

I would like to understand how these folders were used in the UI.

It would be great if we could all use the same generations (`tests/assets`) in the UI as we will use in the "backend" to guide iterations.

tests/generated_artifacts/test_generated_traces.py
```diff
@@ -0,0 +1,32 @@
...
+def generated_trace(request: pytest.FixtureRequest) -> AgentTrace:
+    return AgentTrace.model_validate_json(request.param.read_text())
+
+
+def test_any_tool_used(generated_trace: AgentTrace):
+    assert any(span.is_tool_execution() for span in generated_trace.spans), "No tools were used"
+
+
+@pytest.mark.parametrize("max_steps", range(5, 30, 5))
+def test_steps_taken(generated_trace: AgentTrace, max_steps: int):
```
daavoo, line 22: These are just a couple of examples on how to implement a test with different thresholds and combine it with `xfail`. See #88 of an example using this to verify improvements

aittalam:
> `pytest -v tests/integration/test_single_turn_generation.py --update-assets`

`s/integration/generation/g`, is that correct?

daavoo:
> > `pytest -v tests/integration/test_single_turn_generation.py --update-assets`
> 
> `s/integration/generation/g`, is that correct?

I have corrected the description

khaledosman:
should we add the new output paths to gitignore?

src/agent_factory/generation.py
HareeshBahuleyan: Thanks @daavoo -- removing the filesystemMCP usage simplifies a lot the code

src/agent_factory/generation.py
```diff
@@ -163,51 +97,32 @@ def build_run_instructions(user_prompt) -> str:
         return user_prompt_instance.amend_prompt(user_prompt)
 
 
-def save_agent_trace(agent_trace, latest_dir):
-    agent_trace_path_latest = latest_dir / "agent_factory_trace.json"
-    with Path.open(agent_trace_path_latest, "w", encoding="utf-8") as f:
-        f.write(agent_trace.model_dump_json(indent=2))
-
-
-def archive_latest_run_artifacts(latest_dir, archive_dir):
```
HareeshBahuleyan, line 172: I liked the previous approach of having the `latest` agent inside of folder to be able to find and run easily. And everything else in `archive`. Not sure if we should get rid of that.

tests/conftest.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+
+def pytest_addoption(parser: pytest.Parser) -> None:
+    parser.addoption(
+        "--update-assets",
```
HareeshBahuleyan, line 6: (Minor) Should we call them `artifacts` instead of `assets`?

HareeshBahuleyan (commented):
Thanks a lot for the first set of tests @daavoo ! 

The tests/ part looks great to me. Added a few questions regarding the changes in `src/agent_factory/generation.py`

src/agent_factory/generation.py
```diff
@@ -163,51 +97,32 @@ def build_run_instructions(user_prompt) -> str:
...
-        if item.is_file():
-            shutil.copy(item, archive_dir / item.name)
-        else:
-            print(f"Skipping directory: {item.name}")
-
-
-def single_turn_generation(user_prompt: str, workflow_dir: Path | None = None):
+def single_turn_generation(
+    user_prompt: str,
+    output_dir: Path | None = None,
```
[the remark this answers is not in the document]
aittalam: Yep, I think this works for tests because the `tmp_path` variable is already defined as a `Path`. Suggestion below

src/agent_factory/tools.py
```diff
@@ -38,3 +36,26 @@ def search_mcp_servers(keyword: str, is_official: bool = False) -> list[dict[str
...
+
+    Raises:
+        ValueError: For the following cases:
+            - If the path to the file is not allowed.
+    """
+    file_path = Path(file_name)
+
+    # TODO: this is just a hacky way to restrict file access to
+    # "mimic" the MCP filesystem server.
+    if file_path.parent.name != "tools":
```
aittalam, line 58: NIT: Would it make sense to restrict this even more, let's say to something like `Path.cwd() / "tools"`?

tests/generated_artifacts/test_generated_agents.py
```diff
@@ -0,0 +1,20 @@
...
+@pytest.fixture(
+    params=list(ASSETS_PATH.rglob("*agent.py")),
+    ids=lambda x: x.parent.name,
+)
+def generated_agent_code(request: pytest.FixtureRequest) -> str:
+    return request.param.read_text()
+
+
+def test_specific_tool_used(generated_agent_code: str, request: pytest.FixtureRequest):
+    if "url-to-podcast" in request.node.callspec.id:
```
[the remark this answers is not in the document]
aittalam: Thank you! I think it is great to have one such example, then we can customize it for other more practical use cases 🙏

aittalam (commented):
Many thanks for this PR @daavoo , I think this is going to be very useful to us!

Overall I am more than ok with it, I have just added a couple of comments / suggestions, mostly small things, except the default folder naming when an output path is not provided which I believe might address the issue raised by @khaledosman . 

For the records, I understand that to make decisions re: these non-deterministic agents we should repeat the tests multiple times, but we have already talked about this in our meeting and I think your suggestion of using pytest features to repeat the tests is great. For the sake of simplicity, I would merge this PR as-is, then tackle the multiple runs in a new issue.

src/agent_factory/generation.py
daavoo: @HareeshBahuleyan could you explain a little bit the motivation of `archive`/`latest` ? For example, how was it used?

I was trying to simplify as much as possible by just making `generation` not "opinionated" and just allow an output_dir to be passed. 

I feel like if someone wants a specific output structure you could easily implement it on top of this by passing the `output_dir` you want.

Whereas having the `artifact`/`latest` implemented inside here makes `generation` just less flexible (i.e. I would need to hack my way around to use it in the tests)

tests/conftest.py
daavoo: I think I will just rename everything to `artifacts` 👍

src/agent_factory/tools.py
daavoo: TBH, I didn't iterate much on this because I wanted to open an issue to discuss a potential change to stop relying on reading `tools` from a local folder.

My theory is that the agent doesn't really need to `read_file` the local tools and that we should instead provide it with a description and make the tools a separate package and/or part of `agent-factory` package.

This would have the following benefits:

- Agent doesn't need to read local files to figure out what tools are available.
  Instead, we either inject the available tools directly in the prompt or provide a tool like `list_local_tools` that returns something similar to `search_mcp_servers`, enough information for the agent to understand what are the tools and what they do.
- Generated agent doesn't rely in a local import. 
  Instead, we make `agent-tools` or `agent-factory` a mandatory dependency of the generated agents, so the imports `from something import tool` will work without being inside the `agent-factory` repo.

src/agent_factory/generation.py
daavoo: > The main concern that comes to mind is a security one, for example the LLM can be asked to read and write credentials file like .env files or an ~/.aws/credentials etc.. versus perhaps if its run as a local MCP in docker we have more control over which parts of the system it has access to and can also restrict memory, etc.. as opposed to perhaps an attacker that could try to read/write a huge file or whatever

Agree it is a valid concern. I am not against adding the MCP back if there is need for it.

The thing here is that we "own" the agent factory agent and what tools it has available, so we can apply any restrictions we want. In fact, the way the MCP was set up it was allowing the agent to take more actions than needed, compared to just providing it with a `read_file` tool where you are in full control of the logic.

src/agent_factory/tools.py
daavoo: https://github.com/mozilla-ai/agent-factory/issues/95

src/agent_factory/generation.py
HareeshBahuleyan: >Whereas having the artifact/latest implemented inside here makes generation just less flexible (i.e. I would need to hack my way around to use it in the tests)

Understood, thanks for the clarification. 

Right, there are 2 distinct uses of the generation:
1. For the agent-factory **end-user** to run the generated agent
2. For the agent-factory **developer** to use the generated agent in the tests

I think this refactoring makes it easier for 2. 

However for 1., the end-user needs to specify the output_dir each time (if they don't want to have the time-stamped UUID one). 
With the `latest/`, it was always the same place to run from: `python generated_workflows/latest/agent.py arg1`
A small extra step, but I think its okay. We can proceed with your changes.

Note: At some point, we would need to store the code and artifacts for the end-user to some database and then we have a platform to run the generated agent, but for now its still the file system 😢

tests/assets/url-to-podcast/INSTRUCTIONS.md
```diff
@@ -0,0 +1,33 @@
+### Setup Instructions
```
dpoulopoulos, line 1: I would rename this file to just README.md instead of INSTRUCTIONS.md.

tests/generated_artifacts/test_generated_agents.py
```diff
@@ -0,0 +1,20 @@
+from pathlib import Path
```
dpoulopoulos, line 1: Do we really need separate directories for the tests? Does it simplify the CI runs? What if we had this structure:

- tests/
  - assets/
  - test_generation.py
  - test_generated_agent.py
  - test_generated_trace.py
  - conftest.py
  
Would this work?

src/agent_factory/tools.py
aittalam: I love the idea of having agent-builder tools in a package (see my message on slack yesterday :-))

src/agent_factory/generation.py
khaledosman: IMO I think `latest` is problematic for the following reasons:

1. It makes it difficult to run on cloud or scale; multiple agents being generated at the same time would overwrite each other or try to write to the same place.
2. It has a different path than other agents when trying to read them which can be annoying to deal with, i.e `generated_workflows/latest/agent.py` vs `generated_workflows/archive/workflow_x/agent.py` which is one level deeper so you can't just rely on the workflowId being the only variable so it can be tricky reading/writing files in `latest` vs `archive/workflowId` so I often have to do things like this in the NodeJS api `const workflowName = workflowPath.startsWith('archive/')
      ? workflowPath
      : 'latest'`
3. everything kinda writes to latest by default whether we're generating the agent, running it and generating the trace, the eval cases, the eval results, etc.. it all writes to latest by default, so we can easily end up with a mess with left over files from previous runs... we're not currently clearing everything from latest before generating a new agent afaik

Although I don't have a strong opinion, whatever structure we settle with I should be able to update the NodeJS backend and ui accordingly

tests/assets/url-to-podcast/INSTRUCTIONS.md
daavoo: I think this name was already set before the PR. we could open an issue to discuss

tests/generated_artifacts/test_generated_agents.py
daavoo: >  Does it simplify the CI runs? What if we had this structure:

A little, but we could also just use `-k` flag.

> Would this work?

It would work. I just wanted to somehow make it clearer that there is one where you want to use the `--update-artifacts` flag (`generation`) and others that run on the generated artifacts

src/agent_factory/generation.py
aittalam: Good points @khaledosman and thanks @daavoo !

I am happy with the current code and some documentation explaining how the different options work, and if it is more convenient for anyone to have a `latest` dir for the latest run to play with, we can always explain how to symlink with something like `cd generated_workflows && ln -sfn "archive/$(ls -1t --color=never archive/ | head -1)" latest`

tests/assets/url-to-podcast/INSTRUCTIONS.md
aittalam: Agree, I'd keep this separate from this PR (even just for tracking purposes)

src/agent_factory/generation.py
daavoo: re-requested review. If there is something blocking, please tell, if not, I guess we can move the output structure discussion to a separate issue and/or wait to see what #90 requires

aittalam (approved):
LGTM, thank you @daavoo !

HareeshBahuleyan (approved):

khaledosman (approved):
Pushed [a commit](https://github.com/mozilla-ai/agent-factory/pull/86/commits/05a7df912bd88317c6faf0c483db4b9e8bc26807) to fix the breaking changes affecting the ui & api

Good to go from my side 🚀 

ps: `agent_eval_trace.json` still gets written in `latest/` so I'm keeping a hack in the backend api to copy it over to the workflow directory

Issue #73, opened 2025-06-12
Add Trace-based evaluation to the test suite


stefanfrench:
**Problem statement:**
Currently, we have limited systematic evaluation of the 'creator agent' improvements and regressions when making changes to the codebase instructions. This creates some risks, like regressions that may impact our ability to create a reliable solution.

**Proposed solution:**
Implement trace-based evaluation in the test suite (e.g. test_runner) that can more systematically track the 'creator agent's' performance across iterations. Use any-agent evaluation to do so.

ividal:
Things that will help here:
* there is an evaluation code ready to be used;
* whenever an agent is created, a trace for the manufacturing/parent/x agent is saved (`generated_workflows/latest/agent_factory_trace.json`);
* whenever the created/child/y agent is run, a trace is also saved  (`generated_workflows/latest/agent_eval_trace.json`);
* we have a couple of sample prompts known to be working with [current main (56d62d)](https://github.com/mozilla-ai/agent-factory/tree/56d62d375cf1c34499625f9f6c9b7a6b4f332f7e) in #55 (other complex-but-working-e2e prompts welcome).

Meaning, this issue would easily just cover integrating those pieces together. Ideally these would be part of the CI (although this won't be cheap, so perhaps we want to still keep it to manual triggering?).

javiermtorres:
> Use any-agent evaluation to do so.

Is there a strong reason to perform evaluation in any-agent? Even if both the creator agent and the created agent are any-agent, the evaluation of a process is usually performed "out of band", and there may be other elements involved (MCP servers spring to mind).

PR #88, opened 2025-06-18
tools(search_mcp_servers): Clean up server info and revisit instructi…

daavoo:
…ons.

- The default return contains unnecessary information.
- The available tools are already listed in the search results. No need to manually visit webpage.

---

This is an example to showcase how a change would be introduced after #73 is merged:

- Added a "software test" for the code changes.
- Re-generate assets with `pytest -v tests/generation/test_single_turn_generation.py --update-assets`.
  Verifies that generations are valid.
- Verify improvements with `pytest -v tests/generated_artifacts/`

You can verify in the results that:

- More tests are passing (because token usage has been significantly reduced).
- No new failures were introduced.

| #73 | This PR |
|--------|--------|
| image | image |


src/agent_factory/tools.py
```diff
@@ -35,7 +48,7 @@ def search_mcp_servers(keyword: str, is_official: bool = False) -> list[dict[str
         official_servers = filter(lambda server: server.get("is_official", False), servers)
         return list(official_servers)
```
dpoulopoulos, line 49: Should we also cleanup the official servers?

daavoo: nice catch. I didn't notice that 😅 .

Will update the code and test

aittalam (approved):
Thanks @daavoo! I waited to see that the any-agent version was pinned to make sure the artifacts were regerated with the latest version of the code 😄 , I now think it is ok! I'd trust more @dpoulopoulos for the mcp server stuff, but if that's ok for him I'd say this is ready to merge

dpoulopoulos (approved):

PR #107, opened 2025-06-20
Fix chainlit app

HareeshBahuleyan:
Chainlit app was not working after the changes in #86 and #68

### DIFF
--- a/src/agent_factory/chainlit_app.py
+++ b/src/agent_factory/chainlit_app.py
@@ -109,8 +109,10 @@ async def on_message(message: cl.Message):
 
             step.output = "Running agent with your request..."
 
-            # Run the agent
-            agent_factory_trace = agent_factory.run(task, max_turns=30)
+            # Run the synchronous agent function in a separate thread
+            # This prevents blocking the main asyncio event loop.
+            agent_factory_run = cl.make_async(agent_factory.run)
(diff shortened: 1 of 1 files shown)


dpoulopoulos (approved):

pyproject.toml
```diff
@@ -60,3 +59,6 @@ tests = [
   "pytest-timeout>=2.4.0",
   "syrupy>=4.9.1",
 ]
+chainlit-demo = [
+  "chainlit>=2.5.5",
+]
```
njbrake, line 64: Did you file a bug report with Chainlit?

HareeshBahuleyan: Nop, but I saw the issue you mentioned on #120 
It works with the current versions and any-agent 0.20
But the issue is probably when bumping up any-agent.

Then the package conflicts become an issue and chainlit maintainers do not upgrade their dependencies.
If possible, lets keep it for now, as its a good medium to demo some functionality (till we have multi-turn on Khaled's UI atleast).
And yes, Chainlit will be dropped at some point.

njbrake: I guess I hope "chainlit maintainers do not upgrade their dependencies" isn't true? The PR I opened in chainlit was approved but not merged yet 😬 https://github.com/Chainlit/chainlit/pull/2238

PR #108, opened 2025-06-23
Saving partial agent trace when error encountered during agent.run()

### LINKED ISSUE #20: Make generated agent save partial agent_trace on the event of failure during execution
@HareeshBahuleyan - My other thought: can we still output the `agent_trace.json` even if `generated_workflows/agent.py <arg1>` errors/fails? I could see it still being useful to evaluate the agent trace even if it didn't get to the final output (e.g. hits max_turn = 10 or something)

_Originally posted by @stefanfrench in https://github.com/mozilla-ai/agent-factory/issues/7#issuecomment-2912556593_
            
Currently what I get from the any-agent team is that this functionality does not exist and will have to be implemented as a feature in any-agent tracing.

HareeshBahuleyan:
Saving partial agent trace is now enabled by new feature in any-agent >=0.20: https://github.com/mozilla-ai/any-agent/issues/368 

### Changes in this PR:
1. **Refactoring of `single_turn_generation`**, breaking it down into smaller functions
- `setup_output_directory`, `run_agent`, and `save_agent_outputs` (which are executed in sequence)
- When `AgentRunError` is raised, we can extract the partial trace from the error object (useful for example when the agent fails when it reaches max_turns)
- The refactoring also helps with unit testing of individual functionalities (see below)
- Modified the agent-factory instructions to use AgentRunError and partial trace retrieval in the generated agent code

2. **Tests**
- New unit tests added: `tests/generation/unit/test_generation_utils.py` to cover:
  - Output directory setup logic,
  - Agent run (including error and partial trace scenarios).
- Added a test in tests/generated_artifacts/test_generated_agents.py to ensure that the **generated agent code** includes proper handling for partial traces.

Note: For review, you may ignore the new artifacts for url-to-podcast (`tests/artifacts/`), which were re-generated! 

Closes #20

### DIFF
--- a/tests/generation/unit/test_generation_utils.py
+++ b/tests/generation/unit/test_generation_utils.py
@@ -9,63 +9,64 @@ from agent_factory.generation import (
 )
 
 
-class TestSetupOutputDirectory:
-    def test_with_none_creates_unique_directory(self, tmp_path):
-        """Ensures that passing None as output_dir creates a unique directory."""
-        with patch("agent_factory.generation.Path.cwd", return_value=tmp_path):
-            with patch("agent_factory.generation.datetime") as mock_datetime:
-                mock_datetime.now.return_value.strftime.return_value = "2025-01-01_12:00:00"
-                with patch("agent_factory.generation.uuid.uuid4") as mock_uuid:
-                    mock_uuid.return_value.hex = "1234567890abcdef"  # pragma: allowlist secret
+def test_setup_output_directory_with_none_creates_unique_directory(tmp_path):
+    """Ensures that passing None as output_dir creates a unique directory."""
+    with patch("agent_factory.generation.Path.cwd", return_value=tmp_path):
+        with patch("agent_factory.generation.datetime") as mock_datetime:
+            mock_datetime.now.return_value.strftime.return_value = "2025-01-01_12:00:00"
+            with patch("agent_factory.generation.uuid.uuid4") as mock_uuid:
+                mock_uuid.return_value.hex = "1234567890abcdef"  # pragma: allowlist secret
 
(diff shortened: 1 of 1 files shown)


src/agent_factory/generation.py
```diff
@@ -112,21 +109,59 @@ def single_turn_generation(
...
+    except Exception as e:
+        raise RuntimeError(f"Unexpected error during agent execution. Failed to get agent trace: {e}") from e
+
+
+def save_agent_outputs(agent_trace: AgentTrace, output_dir: Path) -> None:
+    # Save the full trace
+    trace_path = output_dir / "agent_factory_trace.json"
+    trace_path.write_text(agent_trace.model_dump_json(indent=2))
+
+    # Validate and save structured outputs from agent_trace.final_output
```
daavoo, line 131: These comments feel unnecessary as they just explain `what/how` (adding no additional info compared to the function names) instead of `why`

daavoo (approved):
Some cosmetic comments, none are blockers

tests/generation/unit/test_generation_utils.py
```diff
@@ -0,0 +1,85 @@
...
+import pytest
+from any_agent import AgentRunError, AgentTrace
+
+from agent_factory.generation import (
+    run_agent,
+    setup_output_directory,
+)
+
+
+class TestSetupOutputDirectory:
```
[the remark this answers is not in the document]
HareeshBahuleyan: Done

src/agent_factory/generation.py
HareeshBahuleyan: You're right. Dropped those comments.

PR #112, opened 2025-06-23
Tests to check if generated requirements file is valid and installable

HareeshBahuleyan:
Some additional tests such as:
* Verifies the first line of requirements.txt matches any-agent[all]=={version}
* Use regex to detect MCP tool usage which is based on `uvx` in agent.py - if yes, check if 'uv' is in requirements.txt. Additionally, ensure `uv` is not present when uvx MCP is not used

### DIFF
--- a/tests/generation/test_single_turn_generation.py
+++ b/tests/generation/test_single_turn_generation.py
@@ -1,4 +1,6 @@
 import ast
+import subprocess
+import tempfile
 from importlib.util import module_from_spec, spec_from_file_location
 from pathlib import Path
 from shutil import copytree
@@ -18,6 +20,47 @@ def _assert_agent_syntax(agent_file: Path):
     ast.parse(agent_file.read_text(encoding="utf-8"))
 
 
+def _assert_requirements_installable(requirements_path: Path, timeout: int = 180):
+    """Verify that the requirements can be installed in a clean environment.
+
+    Args:
+        requirements_path: Path to the requirements.txt file
+        timeout: Maximum time in seconds to wait for installation
+    """
+    with tempfile.TemporaryDirectory() as temp_dir:
+        env_dir = Path(temp_dir) / ".venv"
+
+        try:
+            # Create a new virtual environment
+            subprocess.run(
+                ["uv", "venv", str(env_dir), "--python=python3.11"],
+                check=True,
+                capture_output=True,
+                text=True,
+                timeout=timeout,
+            )
+
+            # Install requirements
+            result = subprocess.run(
+                ["uv", "pip", "install", "-r", str(requirements_path)],
+                check=False,
+                cwd=temp_dir,
+                capture_output=True,
+                text=True,
+                timeout=timeout,
+            )
+
+            assert result.returncode == 0, (
+                f"Failed to install requirements:\nSTDOUT: {result.stdout}\nSTDERR: {result.stderr}"
+            )
+        except subprocess.TimeoutExpired:
+            raise AssertionError(f"Requirements installation timed out after {timeout} seconds") from None
+        except subprocess.CalledProcessError as e:
+            raise AssertionError(
+                f"Failed to create virtual environment:\nSTDOUT: {e.stdout}\nSTDERR: {e.stderr}"
+            ) from e
+
+
 @pytest.mark.parametrize(
     ("prompt_id", "prompt"),
     [
@@ -35,7 +78,11 @@ def test_single_turn_generation(tmp_path: Path, prompt_id: str, prompt: str, req
 
     _assert_generated_files(tmp_path)
 
+    # Verify the generated agent.py has valid Python syntax
     _assert_agent_syntax(tmp_path / "agent.py")
+
+    # Verify requirements can be installed in a clean environment
+    _assert_requirements_installable(tmp_path / "requirements.txt")
 
     update_artifacts = request.config.getoption("--update-artifacts")
 


tests/generation/test_single_turn_generation.py
```diff
@@ -18,6 +21,108 @@ def _assert_agent_syntax(agent_file: Path):
     ast.parse(agent_file.read_text(encoding="utf-8"))
 
 
+def _assert_requirements_format(requirements_path: Path):
```
HareeshBahuleyan, line 24: Nit: Suggestion to rename to be more specific `_assert_requirements_first_line_matches_any_agent_version`

tests/generation/test_single_turn_generation.py
```diff
@@ -18,6 +21,108 @@ def _assert_agent_syntax(agent_file: Path):
...
+    # Pattern to match any-agent[all]==X.Y.Z format
+    pattern = r"^any-agent\[all\]==\d+\.\d+\.\d+.*$"
+
+    if not re.match(pattern, first_line):
+        raise AssertionError(
+            f"First line of requirements.txt should match 'any-agent[all]==VERSION' format. Found: '{first_line}'"
+        )
+
+
+def _assert_requirements_installable(requirements_path: Path, timeout: int = 180):
```
HareeshBahuleyan, line 43: Since we already have 3 functions specifically for testing requirements.txt, maybe we could move them to a separate python file, and simply import it here in `test_single_turn_generation.py`

tests/generation/test_single_turn_generation.py
```diff
@@ -18,6 +21,108 @@ def _assert_agent_syntax(agent_file: Path):
...
+    content = requirements_path.read_text(encoding="utf-8").strip()
+    lines = content.split("\n")
+
+    if not lines:
+        raise AssertionError("requirements.txt is empty")
+
+    first_line = lines[0].strip()
+
+    # Pattern to match any-agent[all]==X.Y.Z format
+    pattern = r"^any-agent\[all\]==\d+\.\d+\.\d+.*$"
```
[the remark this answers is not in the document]
agpituk: Done!

src/agent_factory/instructions.py
```diff
@@ -136,11 +136,14 @@ def run_agent(url: str):
 1. agent_code should contain all the code implementation of the agent which will correspond to the runnable agent.py script
 2. run_instructions should contain clear and concise setup instructions:
     - Environment variables: Instruct the user to create a .env file to set environment variables; specify exactly which environment variables are required
+    - Instructions to install Python package manager uv:
```
daavoo, line 139: how does the agent know which OS is the target?

tests/generation/requirements_validators.py
```diff
@@ -0,0 +1,115 @@
+import re
+import subprocess
+import tempfile
+from importlib.metadata import version
+from pathlib import Path
+
+
+def assert_requirements_first_line_matches_any_agent_version(requirements_path: Path):
```
aittalam, line 8: Isn't first line a bit too restrictive? I know we are asking it explicitly in instructions, but if the factory built a requirements.txt file with the import in another line that would be still correct, right?

I do not think this needs to be fixed in this PR (as this includes both this code and the instructions, and stuff currently works), but I wanted to raise this because perhaps in the longer term we might want to be less strict with the model when it is not necessary.

tests/generation/requirements_validators.py
```diff
@@ -0,0 +1,115 @@
...
+            f"Found: '{first_line}'\n\nFull requirements.txt content:\n{content}"
+        )
+
+
+def assert_mcp_uv_consistency(agent_file: Path, requirements_path: Path):
+    """Verify MCP tool usage consistency with uv dependency."""
+    agent_content = agent_file.read_text(encoding="utf-8")
+
+    # Check for uvx usage using regex (looking for uvx command usage)
+    uvx_pattern = r"uvx\s+"
```
aittalam, line 34: If we are specifically targeting `uvx` inside the MCP configs for tools, we might want to remove the `\s+` (or change it to `\s*`) as it ofter appears as `"uvx"` when it is called with separate parameters, e.g.

```
TOOLS = [
    MCPStdio(
        command="uvx",               # recommended launcher for kagimcp
        args=["kagimcp"],
        env=...
```

`\s+` always looks for at least one extra whitespace after `uvx`, so would not return a match in that case.

tests/generation/requirements_validators.py
```diff
@@ -0,0 +1,115 @@
...
+    uvx_matches = re.findall(uvx_pattern, agent_content)
+    has_uvx_usage = bool(uvx_matches)
```
aittalam, line 36: nit: perhaps `uses_uvx` and `requires_uvx` are a bit more readable? :-)

tests/generation/requirements_validators.py
```diff
@@ -0,0 +1,115 @@
...
+
+            # Install requirements
+            result = subprocess.run(
+                ["uv", "pip", "install", "-r", str(requirements_path)],
+                check=False,
+                cwd=temp_dir,
+                capture_output=True,
+                text=True,
+                timeout=timeout,
+            )
```
aittalam, line 100: Would it be easier to run things in a throwaway env with something like `uv run --with-requirements generated_workflows/latest/requirements.txt --python 3.11 python -V` ? We just ask python version but (1) it will run uv to install everything in a volatile env and (2) it will succeed only if all specified packages are installable

src/agent_factory/instructions.py
aittalam: Good point: I think in this case the expectation is these lines will both end verbatim in the README file, but we cannot foresee how the model will interpret these instructions. Perhaps we can be more explicit here?

tests/generation/requirements_validators.py
aittalam: image

src/agent_factory/instructions.py
HareeshBahuleyan: > these lines will both end verbatim in the README file

This is the behaviour that I have observed, but good point - probably we need to be explicit with the instructions to add both.

tests/generation/requirements_validators.py
HareeshBahuleyan: You are right that it is restrictive. However, I would suggest to keep it since it is a test to validate proper instruction following by the manufacturing agent.

tests/generation/requirements_validators.py
HareeshBahuleyan: Thanks @aittalam ! You 've some nice uv tips and tricks in your pocket 😄

tests/generation/requirements_validators.py
agpituk: Done!

tests/generation/requirements_validators.py
agpituk: Done!

aittalam (approved):
heeey many thanks both for the PR and for addressing my comments so promptly, LGTM! 🙏

Issue #83, opened 2025-06-17
Save evaluation_case.yaml and agent_eval_trace.json in appropriate archive folders


stefanfrench:
**Problem Statement:**
Currently, `evaluation_case.yaml` and `agent_eval_trace.json` are saved to the latest/ folder, regardless of whether they are the latest agent or an archived agent, which makes it hard to track evaluation data per agent/version.

**Task:**
Update the agent run and evaluation scripts to save evaluation_case.yaml and agent_eval_trace.json to the correct archive folder for each agent (e.g.,` /archive/{workflow_id}/v1/)` instead of latest/. (depending on where they are run from)

khaledosman:
ps: I believe the script to generate `evaluation_case.yaml` is already configurable with the workflow path.

```
 python eval/main.py generated_workflows/archive/workflowId
```

already saves it in the right directory

HareeshBahuleyan:
PR #122 resolves the part about saving `agent_eval_trace.json`

For `evaluation_case.yaml` - it may work specifying the `workflowId` as Khaled mentioned but it is non-deterministic as the [script](https://github.com/mozilla-ai/agent-factory/blob/main/eval/generate_evaluation_case.py) relies on Filesystem MCP. Moreover, the [eval instructions](https://github.com/mozilla-ai/agent-factory/blob/main/eval/instructions.py) still contain mentions of `latest/` folder which could cause issues. 

Hence, re-opening that issue to address the remaining part in another PR.

HareeshBahuleyan:
>non-deterministic as the [script](https://github.com/mozilla-ai/agent-factory/blob/main/eval/generate_evaluation_case.py) relies on Filesystem MCP.

New issue to track this: https://github.com/mozilla-ai/agent-factory/issues/166

>the [eval instructions](https://github.com/mozilla-ai/agent-factory/blob/main/eval/instructions.py) still contain mentions of latest/ folder which could cause issues.

Closing as it was addressed in https://github.com/mozilla-ai/agent-factory/pull/163

PR #113, opened 2025-06-23
refactor: refactor ui & api

khaledosman:
Refactors the existing code to be more reusable and maintainable by:
1. Moving all api calls to service files to make the transition to a new api easier
2. Replace references of "workflowPath" based on the generated workflow path in the file system with "workflowId"
3. Create smaller reusable ui components for common ui functionality to make the main components smaller
4. Create helper files for duplicated logic like streaming api responses, error handling, score calculation
5. Create constant files for api endpoints so they're easily identifiable or swappable (missing a transformation layer)
6. Create composables for shared, reusable or complex computed logic
7. Create constants for query keys to avoid potential errors when clearing the cache or reusing query keys
8. refactor api file structure and endpoints
9. move controller logic into services


khaledosman:
Thanks @HareeshBahuleyan, I think none of these are regression issues or unexpected

> 1. I keep seeing this after "Run Agent"
>    image

This is fine, I just added a logging middleware to the backend to log all requests so it appears on the logs, but this is how the frontend figures out if these files exist or not

> 2. I think `generate_evaluation_case.py` is being run with the default `generated_workflow_dir: str = "generated_workflows/latest"` instead of the time stamped folder

This is an issue within agent-factory itself, it always generats the evaluation_case.yaml file in `latest/` directory and its not configurable afaik, so there's a hack in the backend api which copies it over the corresponding workflow dir after generating it. LMK if agent-factory has it configurable now

> 3. When I start the run (or eval) and switch tabs, I lose the streaming message box or the "running" status:

This is also the current behavior as there's no database storing the logs generated for each agent run, it just streams it in realtime, so if you navigate away no longer knows about it

> 4. If the files could be rendered according to their filetype, that would be cool (like as if an IDE). Maybe rendering them as Markdown code blocks might be a quick way to do it (not sure).

Yup, this is the current existing behavior aswell, but definitely something I'm looking to add in the future, got the same feedback from @hasangzl aswell.

HareeshBahuleyan:
Thanks @khaledosman for the clarification and that these are known/expected issues.
Regarding the 2nd point and #83, I have added a comment there and will work on. And ping you once ready.

PR #115, opened 2025-06-24
Test to assert execution time and number of turns within limits for the manufacturing agent

### LINKED ISSUE #114: Assert number of turns and the execution time of the manufacturing agent are within reasonable limits
Add tests that check both the number of turns and the execution time of the manufacuring agent, enabling these tests to be use-case specific (e.g. simple use-case low-execution time, complex use-case can have higher execution time)

HareeshBahuleyan:
The total execution time and the number of turns for the manufacturing agent should be within reasonable limits (on a per use-case basis)

Runs as a part of the `test_single_turn_generation` per use-case.

It is based on the generated artifacts, i.e., the `agent_factory_trace.json` to calculate the actual execution time and the number of turns.

Can be tested by running:
```
pytest -v tests/generation --update-artifacts
```

Closes #114


tests/generation/test_single_turn_generation.py
```diff
@@ -18,25 +20,67 @@ def _assert_agent_syntax(agent_file: Path):
...
+
 @pytest.mark.parametrize(
-    ("prompt_id", "prompt"),
+    ("prompt_id", "prompt", "expected_num_turns", "expected_execution_time"),
     [
         (
             "url-to-podcast",
             "Create a workflow that takes an input web URL and creates an audio podcast with multiple speakers.",
+            30,
+            180,
```
HareeshBahuleyan, line 57: The numbers used here are set arbitrarily.
TBD what thresholds to set for the different use cases.

src/agent_factory/utils/trace_utils.py
```diff
@@ -0,0 +1,19 @@
+from pathlib import Path
+
+from any_agent.tracing.agent_trace import AgentTrace
+
+
+def load_agent_trace(agent_trace_json_file: str | Path) -> AgentTrace:
```
daavoo, line 6: @HareeshBahuleyan because these changes are on generated traces, I think it makes more sense to implement them under `tests/generated_artifacts`, so you don't need to add the `trace_utils` for loading.

tests/generation/test_single_turn_generation.py
```diff
@@ -18,25 +20,67 @@ def _assert_agent_syntax(agent_file: Path):
...
+        expected_execution_time: Maximum allowed execution time in seconds
+    """
+    agent_execution_time_in_seconds = agent_trace.duration.total_seconds()
+    assert agent_execution_time_in_seconds < expected_execution_time, (
+        f"Agent execution time {agent_execution_time_in_seconds:.2f}s exceeded expected "
+        f"threshold of {expected_execution_time}s"
+    )
+
+
+def _assert_num_turns_within_limit(agent_trace: AgentTrace, expected_num_turns: int) -> None:
```
daavoo, line 37: This is kind of already covered by https://github.com/mozilla-ai/agent-factory/blob/6f98d2006acd7448645630c409fe4687f4f7ed09/tests/generated_artifacts/test_generated_traces.py#L21-L25

There different thresholds are tried and you can see improvements by the number of test that pass vs xfail

src/agent_factory/utils/trace_utils.py
HareeshBahuleyan: Yes although I think `load_agent_trace` could become a generic utils function that is loaded else where too (outside of tests). For example, I use it now in eval/
https://github.com/mozilla-ai/agent-factory/blob/3e662d5d85f081ee49521bf449927b163c22154a/eval/run_generated_agent_evaluation.py#L32

PR #118, opened 2025-06-25
Additional use cases to the testing suite

### LINKED ISSUE #103: Add one or more use cases to test suite
**Problem statement:**
Currently tests are limited to a single use cases, reducing confidence in generalizability and coverage across different and more complex agent types

**Task:**
Add one or more use-cases to the test suite, this could be from the `sample_agents/` or otherwise.

HareeshBahuleyan:
Functionality:
- New function `clean_python_code_with_autoflake` to remove unused imports and unused variables generated agent code before saving as `agent.py`.

Tests:
- Added new use cases `summarize-url-content` and `scoring-blueprints-submission`
- MCP related assertions in the generated code
- Elevenlabs/Slack/SQLite related assertions in the generated code
- Updated prompts used in `test_single_turn_generation` to be more specific (e.g. "do not use MCP" - to make the tests more deterministic)

**Note**: The artifacts generated in `tests/artifacts/` can be ignored when reviewing the PR

Closes https://github.com/mozilla-ai/agent-factory/issues/128
Closes #103


tests/generated_artifacts/test_generated_agents.py
```diff
@@ -14,10 +14,23 @@ def generated_agent_code(request: pytest.FixtureRequest) -> str:
...
         assert "extract_text_from_url" in generated_agent_code
         assert "generate_podcast_script_with_llm" in generated_agent_code
+        assert "combine_mp3_files_for_podcast" in generated_agent_code
+        # ElevenLabs MCP related code matching
+        assert "ELEVENLABS_API_KEY" in generated_agent_code
+        assert any(term in generated_agent_code for term in ("MCPStdio", "MCPSse")), (
+            "MCP server(s) required for url-to-podcast workflow"
+        )
+        assert "generate_audio_simple" in generated_agent_code
+        # TODO: how to assert that only subset of MCP server's tools (not more than the required) are used?
```
HareeshBahuleyan, line 33: Any ideas for how to implement this check are welcome :) Thanks!

HareeshBahuleyan: Thanks @aittalam for sharing your thoughts on this. 

> There are two different places we could run this kind of test:
after search_mcp_servers

Do you mean checking the traces to see the parameters (keywords) passed to this function?
I believe that the keyword search is at a high level to filter out the MCP server(s) to be used, but not at the level of the specific tools
Example:
https://github.com/mozilla-ai/agent-factory/blob/0e3c43afe8797ff93299c66535c0b69ee74f9554/src/agent_factory/tools.py#L32

> we already know a set of tools which is relevant after running agent-factory a number of times and we check that each of the MCP tools in the agent code belongs to that set?
Right that is what I had in mind, there are 2 approaches I see:
1. regex on the agent_code string (x tool in string; y,z tools not in string)
2. a more systematic approach with AST parsing (would that be an overkill for this use case?)

> Would our own agent template be helpful to extract exactly the part of the code where we define the MCP tools?
Maybe, not sure. Currently we run the tests on the whole string agent_code string (and not the individual parts of the template - which isnt saved on disk)

aittalam: > Do you mean checking the traces to see the parameters (keywords) passed to this function?
> I believe that the keyword search is at a high level to filter out the MCP server(s) to be used, but not at the level of the specific tools
I was thinking more along the lines of "are we testing how the LLM makes use of the list of tools provided by the search function or are we testing the search function itself, and which one of them do we have more control of?". But I think even focusing on the result down-the-line already provides us some signal, and we can then trace back to the search itself if we see that becomes a problem.

> Right that is what I had in mind, there are 2 approaches I see:
Cool, I think regex with known lib names could be a good starting point. Thanks!

tests/generated_artifacts/test_generated_agents.py
```diff
@@ -14,10 +14,38 @@ def generated_agent_code(request: pytest.FixtureRequest) -> str:
...
+        assert "summarize_text_with_llm" in generated_agent_code
+        assert all(term not in generated_agent_code for term in ("MCPStdio", "MCPSse")), (
+            "MCP server not required for summarize-url-content workflow"
+        )
+    elif "url-to-podcast" in request.node.callspec.id:
         assert "extract_text_from_url" in generated_agent_code
         assert "generate_podcast_script_with_llm" in generated_agent_code
+        assert "combine_mp3_files_for_podcast" in generated_agent_code
+        # ElevenLabs MCP related code matching
+        assert "ELEVENLABS_API_KEY" in generated_agent_code
```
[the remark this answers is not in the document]
aittalam: It does! I shared my more general concern in our team meeting and I am happy with the answer "we know sticking with mcpm is suboptimal but we'll take care of that at a later stage" :-)

tests/generated_artifacts/test_generated_agents.py
aittalam: That is great Hareesh, many thanks!

aittalam (approved):
Many thanks @HareeshBahuleyan for the PR and for addressing my comments!
Looks good to me 👍

Issue #90, opened 2025-06-18
Convert Agent factory into an API


stefanfrench:
**Problem Statement**
Agent-Factory currently CLI-based, with a NodeJS wrapper that enables the front-end. To meet the platform architecture needs, we need to transition Agent-Factory to be API based. 

**Task**
Refactor Agent-factory to expose its functionality via a FastAPI-based backend, and explore a Python library to enable shell-access method.

stefanfrench:
@dpoulopoulos @peteski22 - please feel free to brake this down into sub-issues, as imagine there are multiple part to this

dpoulopoulos:
This is irrelevant since the agent's lifecycle is now handled by the platform.

PR #120, opened 2025-06-25
feat: upgrade to use evaluation primitives

njbrake:
Use "evaluation primitives" instead of black box `evaluate` function.

The current version of these changes removes the custom weighting of criteria. Let me know if you want that back in, not sure if it's used for anything.

- Changed all references from `evaluation_case.yaml` to `evaluation_case.json` in the codebase.
- Updated the evaluation case generation to save as JSON instead of YAML.
- Adjusted tests and UI components to reflect the new JSON format.
- Bumped `any-agent` dependency to version 0.22.0 for compatibility.

### DIFF
--- a/eval/run_generated_agent_evaluation.py
+++ b/eval/run_generated_agent_evaluation.py
@@ -1,24 +1,26 @@
+import asyncio
 import json
 from pathlib import Path
 
 import fire
-from any_agent.evaluation.evaluate import evaluate
-from any_agent.evaluation.evaluation_case import EvaluationCase
+from any_agent.evaluation import AgentJudge
+from any_agent.evaluation.schemas import EvaluationOutput
 from any_agent.tracing.agent_trace import AgentTrace
 
 from agent_factory.logging import logger
+from eval.generate_evaluation_case import JSONEvaluationCase
 
 
-def run_evaluation(
-    evaluation_case_yaml_file: str = "generated_workflows/latest/evaluation_case.yaml",
+async def run_evaluation(
+    evaluation_case_json_file: str = "generated_workflows/latest/evaluation_case.json",
     agent_trace_json_file: str = "generated_workflows/latest/agent_eval_trace.json",
     save_evaluation_results_path: str = "generated_workflows/latest/evaluation_results.json",
 ):
     """Runs the evaluation process based on an evaluation case YAML file and an agent trace JSON file.
 
     Args:
-        evaluation_case_yaml_file (str): Path to the evaluation case YAML file.
-        Defaults to "generated_workflows/evaluation_case.yaml".
+        evaluation_case_json_file (str): Path to the evaluation case JSON file.
+        Defaults to "generated_workflows/evaluation_case.json".
 
         agent_trace_json_file (str): Path to the agent trace JSON file.
         Defaults to "generated_workflows/agent_eval_trace.json".
@@ -28,8 +30,9 @@ def run_evaluation(
     """
     try:
         # Load evaluation case from the specified YAML file
-        evaluation_case = EvaluationCase.from_yaml(evaluation_case_yaml_file)
-        logger.info(f"Successfully loaded evaluation case from: {evaluation_case_yaml_file}")
+        with Path(evaluation_case_json_file).open(encoding="utf-8") as f:
+            evaluation_case = JSONEvaluationCase.model_validate_json(f.read())
+        logger.info(f"Successfully loaded evaluation case from: {evaluation_case_json_file}")
 
         # Load agent trace from the specified JSON file
         with Path(agent_trace_json_file).open(encoding="utf-8") as f:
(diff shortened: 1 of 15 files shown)


njbrake:
When I was upgrading any-agent version there's now an incompatibility with newer versions of chainlit. I filed a bug report (because it's a chainlit issue) but given that they haven't made a release in several months I'm skeptical that my bug will be addressed quickly https://github.com/Chainlit/chainlit/pull/2238. How crucial is chainlit for the project? It may be worth considering removing it from agent-factory: based on the readme in their repo it appears that it may have been abandoned and is searching for someone to help maintain it again.

njbrake:
> I got an error when running the evaluation step image
> 

You need to run `uv sync` (or whatever the pip install command is for this repo) to pull in the new version of any-agent.


> Also the new json format for the evaluation cases seems to be in the format of
> 
> ```
> {
>   "criteria": [
>   "something",
>   "something else"
>   ]
> }
> ```
> 
> Do we no longer want to attach weights or scores to each point?

The current version of these changes removes the custom weighting of criteria. Let me know if you want that back in, not sure if it's used for anything.

khaledosman:
Would be nice to have the custom weighting back as its a breaking change for the ui, I think its a cool feature in general to let users set custom weights for their criteria especially considering they can create their own criteria from the ui

eval/run_generated_agent_evaluation.py
```diff
@@ -1,24 +1,26 @@
...
 
 
-def run_evaluation(
-    evaluation_case_yaml_file: str = "generated_workflows/latest/evaluation_case.yaml",
+async def run_evaluation(
+    evaluation_case_json_file: str = "generated_workflows/latest/evaluation_case.json",
     agent_trace_json_file: str = "generated_workflows/latest/agent_eval_trace.json",
     save_evaluation_results_path: str = "generated_workflows/latest/evaluation_results.json",
 ):
     """Runs the evaluation process based on an evaluation case YAML file and an agent trace JSON file.
```
HareeshBahuleyan, line 19:
```suggestion
    """Runs the evaluation process based on evaluation case and agent trace.
```

eval/run_generated_agent_evaluation.py
```diff
@@ -27,46 +29,35 @@ def run_evaluation(
         Defaults to "generated_workflows/evaluation_results.json".
     """
     try:
-        evaluation_case = EvaluationCase.from_yaml(evaluation_case_yaml_file)
-        logger.info(f"Successfully loaded evaluation case from: {evaluation_case_yaml_file}")
+        # Load evaluation case from the specified YAML file
```
HareeshBahuleyan, line 32:
```suggestion
```

eval/run_generated_agent_evaluation.py
```diff
@@ -27,46 +29,35 @@ def run_evaluation(
...
-            trace=agent_trace,
-        )
-
+        # Perform the evaluation
+        agent_judge = AgentJudge(model_id="gpt-4.1")
+        results: list[EvaluationOutput] = []
+        runs = []
+        for criteria in evaluation_case.criteria:
+            runs.append(agent_judge.run_async(agent_trace, criteria))
+        results = await asyncio.gather(*runs)
```
HareeshBahuleyan, line 46: Nice, I think this makes it faster.

Side note: When I checked the [any-agent docs](https://mozilla-ai.github.io/any-agent/evaluation/), I noted that the async processing is missing. Maybe its helpful to add there for users to discover.

njbrake:
I don't have any testing code to run for the frontend, but I can see that this PR will need an assist from @khaledosman  in order to get the UI updated with whatever changes happen in the python side 👍

eval/run_generated_agent_evaluation.py
njbrake: Thanks! created https://github.com/mozilla-ai/any-agent/issues/537 to track this.

khaledosman (approved):
> At the moment, as I understand it, there is a tight coupling between frontend (JS/VUE/etc) and python (the backend). to make this a scalable system we may need to consider changing the design so that information is passed back and forth via an API instead of relying on file saving and loading. It's up to the architects of agent-factory to decide, but my suggestion of removing the logic about weights to make room for more broad discussions like whether it makes sense to pass a file back and forth at all.

Kind of, there's a NodeJS api which basically reads/parses/writes these files as necessary, and spawns subprocesses to run the python scripts and stream the response to the frontend, but we do want to move towards a python based api and refactor agent-factory code into some kind of sdk so that it can be reused across different entrypoints like a cli and an api in the future, which is what @dpoulopoulos is working on in https://github.com/mozilla-ai/agent-factory/issues/90

Anyway the changes in https://github.com/mozilla-ai/agent-factory/pull/124 should make the api/ and ui/ compatible with the new structure with the caveat of losing the ability to set custom weights or setting different models for llm as a judge as its no longer supported.

njbrake:
Thanks 💯 for all the help and discussion, @khaledosman! @HareeshBahuleyan In that case I think we're all set, can you do the merge into main if you're good with the changes? I hesitate to do the merge myself just in case you have other concerns.

HareeshBahuleyan:
Running the steps from the UI works as expected. Thanks @njbrake and @khaledosman for the changes!
image

PR #122, opened 2025-06-27
Agent trace saved to same directory as agent.py script

### LINKED ISSUE #83: Save evaluation_case.yaml and agent_eval_trace.json in appropriate archive folders

HareeshBahuleyan:
instead of hardcoded latest/ folder

Closes #83

### DIFF
--- a/api/src/services/evaluation.service.ts
+++ b/api/src/services/evaluation.service.ts
@@ -30,9 +30,6 @@ export class EvaluationService {
     }
 
     await processService.runPythonScript(agentPath, [], outputCallback)
-
-    // Copy trace file from latest to workflow directory if needed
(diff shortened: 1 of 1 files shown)


khaledosman (approved):
Works fine now, I pushed a change to remove the hack from the api that copies the trace file to the agent's folder since its not needed anymore

PR #124, opened 2025-06-27
remove scoring, adapt to new structure of evaluation_case & evaluation_results

khaledosman:


njbrake:
Thanks @khaledosman ! 💯

PR #127, opened 2025-06-27
Add costs to the agent factory trace and to the generated agents

### LINKED ISSUE #125: Aggregate costs incurred per agent
A metric to evaluate the factory can be how costly it is to generate a working agent (using a less powerful model might cause more errors and more back and forth, ramping up costs).

Any-agent (through LiteLLM) returns tokens used and costs per interaction. So we should be able to aggregate them and get the $$$/agent :)

agpituk:
Based on this conversation https://mzlai.slack.com/archives/C08U7DUMSE4/p1750432392712559 I'm trying to get the agent cost added to the agent_factory_trace accessing the property: https://mozilla-ai.github.io/any-agent/api/tracing/#any_agent.tracing.agent_trace.AgentTrace.cost

I added at the end of the trace, after the final_output section

```
  "execution_costs": {
    "input_cost": 0.03659,
    "output_cost": 0.015696,
    "total_cost": 0.052286
  }
```

These are added to both the agent factory and to the every new generated agent

Closes #125


src/agent_factory/generation.py
```diff
@@ -91,10 +92,56 @@ def run_agent(agent: AnyAgent, user_prompt: str, max_turns: int = 30) -> AgentTr
         return e.trace
 
 
+def extract_cost_data(agent_trace: AgentTrace) -> tuple[float, float, float]:
+    """Extract cost data from agent trace and return as (input_cost, output_cost, total_cost)."""
+    try:
+        cost_info = agent_trace.cost
```
ividal, line 98: At this stage, is the end total after using N tools?

src/agent_factory/generation.py
```diff
@@ -91,10 +92,56 @@ def run_agent(agent: AnyAgent, user_prompt: str, max_turns: int = 30) -> AgentTr
...
+        print("No cost information available")
+
+
 def save_agent_outputs(agent_trace: AgentTrace, output_dir: Path) -> None:
-    # First save the agent trace
+    # Log cost information to console
+    log_cost_information(agent_trace)
+
+    # Create enriched trace data with costs as separate metadata
+    input_cost, output_cost, total_cost = extract_cost_data(agent_trace)
```
ividal, line 137: How will this work with multi-turn? I assume any-agent handles that and we step in to collect the latest aggregation?

src/agent_factory/instructions.py
```diff
@@ -217,6 +217,19 @@ def run_agent(url: str):
     ),
 )
 
+def extract_cost_data(agent_trace):
```
ividal, line 220: Not blocking: for legibility I would put this in a separate sub-template in this same module.

src/agent_factory/generation.py
```diff
@@ -91,10 +92,56 @@ def run_agent(agent: AnyAgent, user_prompt: str, max_turns: int = 30) -> AgentTr
...
+def log_cost_information(agent_trace: AgentTrace) -> None:
+    """Log cost information from the agent trace."""
+    input_cost, output_cost, total_cost = extract_cost_data(agent_trace)
+
+    if total_cost > 0:
+        logger.info("=== COST TRACKING ===")
+        if input_cost > 0 and output_cost > 0:
+            formatted_cost = f"input_cost=${input_cost:.6f} + output_cost=${output_cost:.6f} = ${total_cost:.6f}"
+            logger.info(f"Total Cost: {formatted_cost}")
+            print(f"Agent execution cost: {formatted_cost}")
```
ividal, line 123: Do you need the extra print on top of the logger?

src/agent_factory/generation.py
```diff
@@ -91,10 +92,56 @@ def run_agent(agent: AnyAgent, user_prompt: str, max_turns: int = 30) -> AgentTr
         return e.trace
 
 
+def extract_cost_data(agent_trace: AgentTrace) -> tuple[float, float, float]:
```
ividal, line 95: When you add tests to this, I would try for a multi-turn to check what I was commenting on earlier.

ividal (commented):
Thanks for this @agpituk ! We had been considering it for a bit :)

Just a few notes here and there.

src/agent_factory/generation.py
agpituk: I am sure, from looking at the [source code](https://github.com/mozilla-ai/any-agent/blob/main/src/any_agent/tracing/agent_trace.py#L354) of any-agent
`if span.is_llm_call():` on line [354](https://github.com/mozilla-ai/any-agent/blob/0daeac076a67fc2d67c2291bcb86c21d009787c6/src/any_agent/tracing/agent_trace.py#L354)

Looks like only `is_llm_call` are being counted and not `is_tool_execution` . Maybe @mozilla-ai/any-agent-team can help us!

src/agent_factory/generation.py
agpituk: I assume the same, but, it would be nice to test. Do we you know if we've got a branch with the multi_turn generation so I can have a look and test it?

src/agent_factory/generation.py
agpituk: Thanks @daavoo ! Trying to understand this properly. For example, when executing the `def search_web` tool. This The costs come from the LLM choosing to use that tool, and from the LLM to process the output of the call, is that correct?
And in that case, are those two already included in the cost traces?

javiermtorres (approved):
LGTM, I don't see any obvious issue 🤔

ividal (approved):
Thanks @agpituk ! 🚢

src/agent_factory/generation.py
ividal: Thanks for the tests!

PR #141, opened 2025-07-03
Added costs tracking for generation of an evaluation case

agpituk:
## 📝 What's changing

Following #127 this PR adds cost tracking to the evaluation case.

It currently output things in the same evaluation file:
`{
  "criteria": [
    "Ensure that the agent performs a Brave web search (using 'search_tavily') to find a relevant, reputable URL when the input is a query and not a direct URL.",
    "Ensure that the agent correctly identifies when the input is already a URL and skips redundant searching in this case.",
    "Check that the agent uses 'extract_text_from_url' to extract the readable main content from the identified URL, omitting navigation, ads, and unrelated elements.",
    "Verify that the agent summarizes the extracted text using 'summarize_text_with_llm', producing a concise paragraph (maximum 200 words or 5 sentences).",
    "Ensure that if the extracted content is very long, the agent prioritizes summarizing the beginning and main sections of the text.",
    "Check that the agent always returns a structured output containing (a) the summarized text ('summary'), (b) the URL that was summarized ('url'), and (c) a brief excerpt from the full extracted text (the first ~200 characters, as 'full_text_excerpt').",
    "Ensure that the agent does not fabricate or hallucinate content and base the summary strictly on the extracted text.",
    "Ensure that if text extraction fails, the agent clearly reports the extraction error instead of outputting a summary.",
    "Verify that the output JSON strictly adheres to the 'StructuredOutput' schema, properly populating 'url', 'summary', and 'full_text_excerpt' fields with accurate data."
  ],

  "evaluation_case_generation_costs": {
    "input_cost": 0.014242000000000001,
    "output_cost": 0.0026239999999999996,
    "total_cost": 0.016866
  }
}`

I'm open to use a different file for this, happy to discuss it!

As soon as https://github.com/mozilla-ai/any-agent/pull/576 gets merged and we use this version of any-agent, I'll also add the costs for the AgentJudge bit.

---

## 📚 How to test it

Steps to test the changes:

1. run an evaluation:  `uv run python -m eval.run_generated_agent_evaluation `
2. the `evaluation_case.json` should have a new section with the costs

## ✅ Pre-merge checklist

- [X] Tested the changes in a working environment to ensure they work as expected
- [ ] [Manually triggered](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow) workflows from this branch and ensured that test that involve agent generation using LLM API keys run successfully

---


eval/generate_evaluation_case.py
```diff
@@ -75,12 +76,21 @@ def main(generated_workflow_dir: str = "generated_workflows/latest"):
...
+        "output_cost": cost_info.output_cost,
+        "total_cost": cost_info.total_cost,
+    }
+
     if not isinstance(agent_trace.final_output, JSONEvaluationCase):
         raise ValueError("The agent's final output is not a JSONEvaluationCase.")
 
-    # Save the JSON output as a json file
+    evaluation_case_data = agent_trace.final_output.model_dump()
+    evaluation_case_data["evaluation_case_generation_costs"] = evaluation_case_generation_costs
```
HareeshBahuleyan, line 90: We might need to double check if introduction of this new field breaks the schema check, because it is being validated in the run_agent_evaluation step:
https://github.com/mozilla-ai/agent-factory/blob/96482c463eddf2df77f3a14ef39baa26398e9229/eval/run_generated_agent_evaluation.py#L34

HareeshBahuleyan: It can be verified by running steps 1 to 4 in https://github.com/mozilla-ai/agent-factory?tab=readme-ov-file#1-generate-the-workflow

agpituk: > I think we may need to update the mock data used in the tests: https://github.com/mozilla-ai/agent-factory/tree/main/tests/generated_agent_evaluation/data Both for this PR and #127

good shout!  added

agpituk: > It can be verified by running steps 1 to 4 in https://github.com/mozilla-ai/agent-factory?tab=readme-ov-file#1-generate-the-workflow

I've ran this locally and everything still works as expected, but I think it makes sense to update the model so I just updated it as well

PR #152, opened 2025-07-04
Add agent-factory backend

dpoulopoulos:
## 📝 What's changing

This PR adds the agent-factory backend, which provides HTTP endpoints to:
* Create agents using the agent-factory agent.
* Store metadata about agents in a database.
* Store the generated agent files on an S3-compatible object storage. For local deployments, use MinIO.
* Retrieve generated agents as `.zip` files
    
The async, long-running process of creating an agent is offloaded to a Celery task, thus, the app needs a broker to run. The chosen broker in this case is Redis.

Refs #90

## Note to reviewers

Summarizing the components and data flow for the backend:

   * `main.py`: The entry point of the FastAPI application. It initializes the database and creates the FastAPI app.
   * `api/routes/agents.py`: Defines the API endpoints for agent-related operations (create, get, list, download).
   * `services/agents.py`: Implements the `AgentService` class, which contains the business logic for managing agents. It uses the `AgentRepository` for database interactions and Celery for asynchronous tasks.
   * `tasks.py`: (Inferred) This module defines the Celery task (`send_message_task`) for sending messages to the A2A agent.

sequence-diagram

agent-factory-backend

Data Flows:
   1. Create Agent:
       * A POST request with the agent's prompt is sent to the `/api/v1/agents/` endpoint.
       * The `create_agent` function in api/routes/agents.py receives the request.
       * It calls the `create_agent` method of the `AgentService`.
       * The [AgentService](url) builds a prompt, creates an agent record in the database using the `AgentRepository`, and dispatches a `send_message_task` to Celery.
       * The `AgentService` returns an `AgentSummary` to the API, which is then sent back to the client.
   2. Get Agent:
       * A GET request with the agent's ID is sent to the `/api/v1/agents/{agent_id}` endpoint.
       * The get_agent function in `api/routes/agents.py` receives the request.
       * It calls the `get_agent` method of the `AgentService`.
       * The `AgentService` retrieves the agent record from the database using the `AgentRepository`.
       * It returns an `AgentConfig` object to the API, which is then sent back to the client.
   3. List Agents:
       * A GET request is sent to the `/api/v1/agents/ `endpoint.
       * The `list_agents` function in `api/routes/agents.py` receives the request.
       * It calls the `list_agents` method of the [AgentService](url).
       * The [AgentService](url) retrieves all agent records from the database using the [AgentRepository](url).
       * It returns a list of [AgentSummary](url) objects to the API, which is then sent back to the client.
   4. Download Agent:
       * A GET request with the agent's ID is sent to the `/api/v1/agents/{agent_id}/download` endpoint.
       * The `download_agent` function in `api/routes/agents.py` receives the request.
       * It calls the `download_agent` method of the `AgentService`.
       * The `AgentService` downloads the agent's files, creates a zip archive, and returns the file path.
       * The API returns a `FileResponse` to the client, allowing them to download the zip file.

Celery Workflow:
   1. Task Dispatch:
       * When a new agent is created via the `create_agent` endpoint, the `AgentService` calls `send_message_task.delay(prompt, agent_id=str(record.id))`.
       * The `.delay()` method is a Celery feature that sends the `send_message_task` to the Redis message broker. This is a non-blocking call, so the `AgentService` can immediately return a response to the client without
         waiting for the task to complete.
   2. Task Execution:
       * A Celery worker, running as a separate process, continuously monitors the Redis broker for new tasks.
       * When the worker picks up the `send_message_task`, it executes the `send_message_task` function defined in `tasks.py`.
   3. `send_message_task` Execution Steps:
       * Update Status to "Processing".
       * Send Message to the A2A Agent calling the `_send_message_async` function, which in turn uses an `AgentClient` to send the prompt to the external "manufacturing agent". This is an asynchronous HTTP request.
       * Handle Response:
           * If the request is successful and the manufacturing agent returns a successful response, the task parses the JSON result.
           * The `_save_agent` function is called to save the agent's generated code, README, and dependencies to an S3 bucket.
           * The agent's status is updated to COMPLETED.

## 📚 How to test it

The best way to test this end-to-end is by using the docker-compose.yaml file. Run docker-compose up and navigate to localhost:8000/docs.

---

## ✅ Pre-merge checklist

- [x] Added some tests for any new functionality
- [x] [Manually triggered](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow) workflows from this branch and ensured that test that involve agent generation using LLM API keys run successfully.


macaab26:
Hey @dpoulopoulos, could you provide a schema/overview or a guide to start reviewing the PR? Thank you very much!

src/backend/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Installation
+
+1. Clone the repository and navigate to the backend's source directory:
+   ```bash
+   git clone https://github.com/mzai/agent-factory.git
+   cd agent-factory/src/backend
+   ```
+
+2. Install the dependencies using `uv`:
```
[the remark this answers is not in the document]
dpoulopoulos: Does it matter? `uv sync` creates a virtual environment, if it doesn't already exist, and installs all the dependencies in this environment. Then, you can activate it to run any subsequent commands.

src/backend/backend/services/agents.py
```diff
@@ -0,0 +1,133 @@
...
+        The method immediately returns a summary of the created agent, which includes
+        the agent's ID, summary, status, and creation timestamp.
+
+        Args:
+            request (AgentCreateRequest): The request containing the agent's prompt.
+
+        Returns:
+            AgentSummary: A summary of the created agent.
+        """
+        summary = request.prompt[:120] + "..." if len(request.prompt) > 120 else request.prompt
```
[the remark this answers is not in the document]
dpoulopoulos: Done: https://github.com/mozilla-ai/agent-factory/issues/158

src/backend/README.md
```diff
@@ -0,0 +1,77 @@
...
+- Python 3.13+
+- [uv](https://github.com/astral-sh/uv)
+- Docker (for running Redis)
+- Redis
+
+### Installation
+
+1. Clone the repository and navigate to the backend's source directory:
+   ```bash
+   git clone https://github.com/mzai/agent-factory.git
```
HareeshBahuleyan, line 24:
```suggestion
   git clone https://github.com/mozilla-ai/agent-factory.git
```

src/backend/backend/api/routes/agents.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@router.get("/{agent_id}/download")
+def download_agent(
+    service: AgentServiceDep,
+    agent_id: UUID,
+) -> FileResponse:
+    file_path = service.download_agent(agent_id)
+    try:
+        return FileResponse(path=file_path, filename=f"{agent_id}.zip", media_type="application/octet-stream")
```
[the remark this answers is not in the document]
dpoulopoulos: I think I have to restructure the way I raise errors in general. I'm onto it.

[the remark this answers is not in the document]
dpoulopoulos: Done.

docker-compose.yaml
```diff
@@ -0,0 +1,52 @@
+services:
+  redis:
+    image: redis:alpine
+    ports:
+      - 6379:6379
```
peteski22, line 5:
```suggestion
      - "6379:6379"
```

docker-compose.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+  backend:
+    build:
+      context: ./src/backend
+      dockerfile: Dockerfile
+    ports:
+      - 8000:8000
```
peteski22, line 12:
```suggestion
      - "8000:8000"
```

docker-compose.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - db_data:/app
+    depends_on:
+      - redis
+
+  agent:
+    build:
+      context: ./src/agent
+      dockerfile: Dockerfile
+    ports:
+      - 8080:8080
```
peteski22, line 23:
```suggestion
      - "8080:8080"
```

docker-compose.yaml
```diff
@@ -0,0 +1,52 @@
...
+      context: ./src/backend
+      dockerfile: Dockerfile
+    ports:
+      - 8000:8000
+    volumes:
+      - db_data:/app
+    depends_on:
+      - redis
+
+  agent:
```
peteski22, line 18: ❓ Do you think it's useful to add the required env vars here:

```yaml
environment:
      FRAMEWORK: ${AGENT_FACTORY_FRAMEWORK:-openai}
      MODEL: ${AGENT_FACTORY_MODEL:-o3}
      OPENAI_API_KEY: ${OPENAI_API_KEY}
      HOST: ${AGENT_FACTORY_HOST:-0.0.0.0}
      PORT: ${AGENT_FACTORY_PORT:-8080}
      LOG_LEVEL: ${AGENT_FACTORY_LOG_LEVEL:-info}
```

They mirror the defaults set in the `src/backend/Dockerfile` but add `OPENAI_API_KEY` which I found I needed to get things working when I tested it out.

src/backend/backend/storage.py
peteski22, line 1: ❓ is the `if/else` to differentiate between `s3` and `minio`? 

❓ Do you think we should add the settings to the `README` so it's documented somewhere?

❓ Should the `storage_client` be configured as a dependency ([for example](https://github.com/mozilla-ai/lumigator/blob/275796923e341d02fc1f8d42926279b9eb8161c1/lumigator/backend/backend/api/deps.py#L47))

src/backend/backend/tasks.py
```diff
@@ -0,0 +1,124 @@
...
+    with session_manager.session() as session:
+        yield AgentRepository(session)
+
+
+def _update_agent_status(agent_id: uuid.UUID, status: str):
+    logger.info(f"Updating agent {agent_id} status to {status}")
+    for repo in _get_agent_repository():
+        repo.update(agent_id, status=status)
+
+
```
peteski22, line 33: ❗ I think we need a function to update the `download_available` when it's ready.

```suggestion

def _update_agent_download(agent_id: uuid.UUID, available: bool):
    logger.info(f"Updating agent {agent_id} download_available to {available}")
    for repo in _get_agent_repository():
        repo.update(agent_id, download_available=available)
        
        
```

src/backend/backend/tasks.py
```diff
@@ -0,0 +1,124 @@
...
+        tools_dir = Path("backend/tools")
+        for tool_file in tools_dir.iterdir():
+            if tool_file.is_file() and (tool_file.stem in agent_code or tool_file.name == "__init__.py"):
+                storage_client.upload_file(
+                    str(tool_file),
+                    bucket_name,
+                    f"{agent_id}/tools/{tool_file.name}",
+                )
+
+        logger.info(f"Agent files saved to bucket {bucket_name}")
```
peteski22, line 81: We weren't ever updating the status of the download to mark it as available.

```suggestion
        logger.info(f"Agent files saved to bucket {bucket_name}")
        _update_agent_download(agent_id, True)
```

src/backend/backend/api/routes/agents.py
```diff
@@ -0,0 +1,48 @@
+from uuid import UUID
+
+from fastapi import APIRouter, status
+from fastapi.responses import FileResponse
```
peteski22, line 4:
```suggestion
from fastapi.responses import FileResponse
from starlette.requests import Request
from starlette.responses import Response
```

src/backend/backend/api/routes/agents.py
```diff
@@ -0,0 +1,48 @@
...
+        AgentNotFoundError: status.HTTP_404_NOT_FOUND,
+    }
+
+
+@router.post("/", status_code=status.HTTP_201_CREATED)
+async def create_agent(
+    service: AgentServiceDep,
+    agent_create_request: AgentCreateRequest,
+) -> AgentSummary:
+    return await service.create_agent(agent_create_request)
```
peteski22, line 24: Since we're not returning the full object, we can add the `Location` header to point to where it can be found.

```suggestion
    agent_create_request: AgentCreateRequest,
    request: Request,
    response: Response,
) -> AgentSummary:
    agent_summary = await service.create_agent(agent_create_request)
    url = request.url_for(get_agent.__name__, agent_id=agent_summary.id)
    response.headers["Location"] = f"{url}"
    return agent_summary
```

peteski22 (commented):
Thanks so much for putting this together @dpoulopoulos ❤️ it looks like a lot went into this, and the diagrams etc. on the PR summary are awesome.

I checked out the branch and tried testing it out (with `docker compose up`), but I had a few issues. I've added some suggestions/questions/etc. to the PR, but I wonder if you could take a look at:

* `api/v1/agents/{agent_id}/download` doesn't seem to work - I get a `500` Internal Server Error.
* Logging doesn't seem to be working as intended inside the various containers when the code calls `logger.info()` etc. I couldn't see any logs other than from libraries that aren't using our logger.
* Appreciate the tests that were added, but I think we might need a lot more coverage to reassure ourselves - I had a quick check when I ran into issues to see if there were related tests but I couldn't find them.

I just marked the review as `Comment` and not `Request Changes` as I don't want to block you, but I think some changes will be needed before it's ready to merge.

src/backend/backend/storage.py
dpoulopoulos: > is the if/else to differentiate between s3 and minio?
Yes. I didn't want to use the MinIO Python client (https://pypi.org/project/minio/) because we're going to use S3 on the platform, thus, I wanted to kill two birds with one stone.

> Do you think we should add the settings to the README so it's documented somewhere?
Sure, I'll add this.

> Should the storage_client be configured as a dependency ([for example](https://github.com/mozilla-ai/lumigator/blob/275796923e341d02fc1f8d42926279b9eb8161c1/lumigator/backend/backend/api/deps.py#L47))
I could add this, yes.

src/backend/backend/records/agents.py
```diff
@@ -0,0 +1,13 @@
...
+
+from backend.records.base import BaseRecord
+from backend.records.mixins import AgentStatusMixin, DateTimeMixin
+
+
+class AgentRecord(BaseRecord, DateTimeMixin, AgentStatusMixin):
+    __tablename__ = "agents"
+
+    summary: Mapped[str]
+    prompt: Mapped[str]
```
HareeshBahuleyan, line 11: I initially though that `prompt` was going to be just what the user submits (And `summary` is the LLM generated version in few-words). But I see it to contain the "amendment" that we include. Was that intentional? 

image

Not a blocker, but we probably need a way to back track and have some versioning on the system prompt and amendment, and store those versions for later reference.

src/backend/backend/utils.py
```diff
@@ -0,0 +1,70 @@
...
+    Args:
+        agent_id (UUID): The unique identifier of the agent.
+
+    Returns:
+        Path: The path to the directory where the files were downloaded.
+    """
+    output_dir = prepare_output_dir(agent_id)
+    bucket_name = settings.S3_BUCKET_NAME
+
+    files_to_download = ["agent.py", "README.md", "requirements.txt", "Dockerfile"]
```
HareeshBahuleyan, line 38: I see that we now include this Dockerfile for the agent. But because the system instructions were written for the cli-agent, the generated README does not have any mention of running via docker (but rather the run instructions with uv instead) 
Just raising it here, but we can address this in a later PR.

PR #185, opened 2025-07-23
A2a client refactor

HareeshBahuleyan:
## 📝 What's changing

This PR breaks up agent_factory/test_client.py (renamed to agent_generation.py) into smaller functions that can be individually tested.

We also introduce a `Makefile` to make running a2a server via docker easier - This will be used later to include different test groups later (test-unit, test-integration, test-evaluation, etc. which could make it easier for devs to run tests against their changes - similar to how we had it in Lumigator) 

---

## 📚 How to test it

Steps to test the changes:

1. `make run`
2. Check A2A is running on http://127.0.0.1:8080/.well-known/agent.json 
3. `uv run agent-factory "Curate best sushi restaurants in the city given by the user"`
4. Check results are saved in `generated_workflows/`

---

## ✅ Pre-merge checklist

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,49 @@
+# Makefile for Agent Factory
+
+# Configuration
+DOCKER_IMAGE = agent-factory
+DOCKER_CONTAINER = agent-factory-a2a
+DOCKER_TAG = latest
+A2A_SERVER_HOST = localhost
+A2A_SERVER_PORT = 8080
+
+# Default target
+.PHONY: help
+help:
+	@echo "Available targets:"
+	@echo "  build           - Build the Docker image for Agent Factory A2A server"
+	@echo "  run             - Run the A2A server in Docker"
+	@echo "  stop            - Stop the running container"
+	@echo "  clean           - Remove Docker containers and images"
+
+# Build the Docker image
+.PHONY: build
+build:
+	docker build --build-arg APP_VERSION=$(shell git describe --tags --dirty 2>/dev/null || echo "dev") -t $(DOCKER_IMAGE):$(DOCKER_TAG) .
+
+# Run the server in Docker
+.PHONY: run
+run: build
+	@if [ ! -f .env ]; then \
+		echo "Error: .env file not found. Please create one with your API keys."; \
+		exit 1; \
+	fi
+	docker run --rm -d \
+		--name $(DOCKER_CONTAINER) \
+		-p $(A2A_SERVER_PORT):8080 \
+		--env-file .env \
+		$(DOCKER_IMAGE):$(DOCKER_TAG)
+	@echo "Server running at http://$(A2A_SERVER_HOST):$(A2A_SERVER_PORT)"
+
+# Stop the running container
+.PHONY: stop
+stop:
+	@docker stop $(DOCKER_CONTAINER) 2>/dev/null || true
+
+# Clean up Docker resources
+.PHONY: clean
+clean: stop
+	@docker rmi $(DOCKER_IMAGE):$(DOCKER_TAG) 2>/dev/null || true
+
+# Default target
+.DEFAULT_GOAL := help
--- a/README.md
+++ b/README.md
@@ -85,34 +85,27 @@ In addition to `host` and `port`, you can also pass the following arguments:
 > Visit the any-agent [documentation](https://mozilla-ai.github.io/any-agent/) for more details on the supported
 > frameworks.
 
-#### Using Docker
+#### Using Makefile
 
-To run the server in a Docker container, follow these steps.
+The Makefile enables you to run the server using Docker.
 
-1.  **Build the Docker image:**
+**Run the server** (this will also build the image if needed):
+   ```bash
+   make run
+   ```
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -0,0 +1,47 @@
+# Makefile for Agent Factory
+
+# Configuration
+DOCKER_IMAGE = agent-factory
+DOCKER_CONTAINER = agent-factory-a2a
+DOCKER_TAG = latest
+A2A_SERVER_HOST = localhost
+A2A_SERVER_PORT = 8080
```
dpoulopoulos, line 8:
```suggestion
A2A_SERVER_LOCAL_PORT = 8080
```

Just to distinguish it from the one the agent in the container listens to.

Makefile
```diff
@@ -0,0 +1,47 @@
...
+# Run the server in Docker
+.PHONY: run
+run: build
+	@if [ ! -f .env ]; then \
+		echo "Error: .env file not found. Please create one with your API keys."; \
+		exit 1; \
+	fi
+	docker run --rm -d \
+		--name $(DOCKER_CONTAINER) \
+		-p $(A2A_SERVER_PORT):8080 \
```
dpoulopoulos, line 33:
```suggestion
		-p $(A2A_SERVER_LOCAL_PORT):8080 \
```

Makefile
```diff
@@ -0,0 +1,47 @@
...
+		--env-file .env \
+		$(DOCKER_IMAGE):$(DOCKER_TAG)
+	@echo "Server running at http://$(A2A_SERVER_HOST):$(A2A_SERVER_PORT)"
+
+.PHONY: stop
+stop:
+	@docker stop $(DOCKER_CONTAINER) 2>/dev/null || true
```
dpoulopoulos, line 40: Why don't we return the output of the `docker stop` command?

Makefile
```diff
@@ -0,0 +1,47 @@
...
+
+.PHONY: clean
+clean: stop
+	@docker rmi $(DOCKER_IMAGE):$(DOCKER_TAG) 2>/dev/null || true
```
dpoulopoulos, line 44: Same here. Why don't we return the output of the `docker rmi` command?

README.md
```diff
@@ -85,34 +85,27 @@ In addition to `host` and `port`, you can also pass the following arguments:
 > Visit the any-agent [documentation](https://mozilla-ai.github.io/any-agent/) for more details on the supported
 > frameworks.
 
-#### Using Docker
+#### Using Makefile
```
dpoulopoulos, line 88: I would keep it as `Using Docker`, because we may want to use the Makefile to run the agent as a process as well.

Makefile
```diff
@@ -0,0 +1,47 @@
...
+	@echo "Available targets:"
+	@echo "  build           - Build the Docker image for Agent Factory A2A server"
+	@echo "  run             - Run the A2A server in Docker"
+	@echo "  stop            - Stop the running agent-factory-a2a container"
+	@echo "  clean           - Remove Docker containers and images"
+
+# Build the Docker image
+.PHONY: build
+build:
+	docker build --build-arg APP_VERSION=$(shell git describe --tags --dirty 2>/dev/null || echo "dev") -t $(DOCKER_IMAGE):$(DOCKER_TAG) .
```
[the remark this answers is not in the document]
dpoulopoulos: Oh, sorry, you are redirecting stderr! Nevermind!

Makefile
HareeshBahuleyan: You 're right - updated to be able to see the success/error messages

Makefile
```diff
@@ -0,0 +1,49 @@
...
+	docker run --rm -d \
+		--name $(DOCKER_CONTAINER) \
+		-p $(A2A_SERVER_LOCAL_PORT):8080 \
+		--env-file .env \
+		$(DOCKER_IMAGE):$(DOCKER_TAG)
+	@echo "Server running at http://$(A2A_SERVER_HOST):$(A2A_SERVER_LOCAL_PORT)"
+
+.PHONY: stop
+stop:
+	@docker stop $(DOCKER_CONTAINER)
```
dpoulopoulos, line 40:
```suggestion
@if [ "$$(docker ps -q -f name=$(DOCKER_CONTAINER))" ]; then \
        docker stop $(DOCKER_CONTAINER); \
        echo "Successfully stopped $(DOCKER_CONTAINER) container"; \
else \
        echo "No $(DOCKER_CONTAINER) container running"; \
	fi
```

This addresses the following scenario:

1. The user first runs `make stop` to stop the container.
1. The user then runs `make clean` to remove the image.
    1. The `clean` target first runs the `stop` target.
    1. The `stop` target now fails since there's no such container.
    1. The whole `clean` target fails.

HareeshBahuleyan: Makes sense. Updated to use `&&`
```
docker stop $(DOCKER_CONTAINER) && echo "Successfully stopped $(DOCKER_CONTAINER) container";
```
to handle the edge case if `docker stop $(DOCKER_CONTAINER)` fails for some reason it will still `echo "Successfully stopped $(DOCKER_CONTAINER) container";`

PR #224, opened 2025-08-06
feat: Enable tracing propagation.

daavoo:
Run `generate_target_agent` and you should see the trace being serialized to a file under `traces` and the loaded JSON returned.

- Added a simple `JsonFileSpanExporter` that dumps the traces in JSON format to a directory.


src/agent_factory/agent_generator.py
```diff
@@ -35,40 +43,44 @@ async def generate_target_agent(
         port: The port for the agent server (default: 8080).
         timeout: The timeout for the request in seconds (default: 600).
     """
-    try:
-        http_client, base_url = await create_a2a_http_client(host, port, timeout)
-        async with http_client as client:
-            resolver = A2ACardResolver(httpx_client=client, base_url=base_url)
-            agent_card = await get_a2a_agent_card(resolver)
+    with tracer.start_as_current_span("generate_target_agent") as span:
+        trace_file = f"0x{trace.format_trace_id(span.get_span_context().trace_id)}.json"
```
daavoo, line 47: Main point is that here we have `trace_id` as the single identifier that will be propagated and present in the A2A and any-agent spans, so that is how you can retrieve it later (I added a simple loading of the JSON in the end)

src/agent_factory/utils/json_exporter.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def export(self, spans) -> None:
+        # File name matches how trace_id will be formatted inside the JSON
+        output_file = self.output_dir / f"0x{format_trace_id(spans[0].context.trace_id)}.json"
+        try:
+            all_spans = json.loads(output_file.read_text())
+        except (json.JSONDecodeError, FileNotFoundError):
+            all_spans = []
+
+        for span in spans:
```
HareeshBahuleyan, line 25: To keep only any-agent traces, this could work?
```
        for span in spans:
            # We don't need a2a server events in traces
            if not span.name.startswith("a2a.server"):
                try:
                    span_data = json.loads(span.to_json())
                except (json.JSONDecodeError, TypeError, AttributeError):
                    span_data = {"error": "Could not serialize span", "span_str": str(span)}

                all_spans.append(span_data)

```

src/agent_factory/__main__.py
```diff
@@ -3,15 +3,29 @@
...
 
 
+trace.set_tracer_provider(TracerProvider())
+span_processor = SimpleSpanProcessor(JsonFileSpanExporter("traces"))
+trace.get_tracer_provider().add_span_processor(span_processor)
+
+StarletteInstrumentor().instrument()
+
+
 async def main(
```
HareeshBahuleyan, line 29: When running agent-factory via docker (https://github.com/mozilla-ai/agent-factory/blob/main/README.md#using-docker), we might need to mount paths? For the client to be able to access it.

HareeshBahuleyan (commented):
Hey @daavoo Thanks a lot for the implementation. 

When I tested, the trace propagation seems to have some issues I think.
I see the logs:
```
  INFO     Trace will be saved to                                agent_generator.py:57
                             0x306f6bdcab8df3ac74a7016de4764418.json
```
But its a different trace_id that the server uses:
```
(agent-factory) (base) ➜  agent-factory git:(propagate-and-save-traces) ✗ ls -lt traces
total 112
-rw-r--r--@ 1 hareeshbahuleyan  staff  27418 Aug  6 16:10 0x1f9b8424000d58e7edbb77c3b59bfd87.json
```

daavoo:
> > Hey @daavoo Thanks a lot for the implementation.
> > When I tested, the trace propagation seems to have some issues I think. I see the logs:

Here is what I see



https://github.com/user-attachments/assets/77e01b9e-8307-4001-b5af-ca5ea29dc584

src/agent_factory/__main__.py
HareeshBahuleyan: Fixed in #232

HareeshBahuleyan (approved):

PR #232, opened 2025-08-08
Improvements to trace propagation with JsonFileSpanExporter

HareeshBahuleyan:
## 📝 What's changing

This PR addresses makes minor improvements to PR #224 
1. `TRACES_DIR` will always be project_root/traces (previously it depended on the relative path from where the script was being run)
2. Bind mount `traces/` when running via docker (so that the client can access the exported traces)

In JsonFileSpanExporter
3. Filter to keep only `any-agent` spans `["call_llm", "execute_tool", "invoke_agent"]`
4. Append file instead of read and overwrite

---

## 📚 How to test it

Steps to test the changes:

1. 

---

## ✅ Pre-merge checklist

---

### DIFF
--- a/docs/user-guide/mcp-servers.md
+++ b/docs/user-guide/mcp-servers.md
@@ -20,7 +20,7 @@ The following table provides a quick overview of tested MCP servers. For detaile
 
 <!-- MCP_SERVERS_TABLE_START -->
 
-*Last updated: 2025-08-08T09:22:37.615813+00:00*
+*Last updated: 2025-08-08T13:02:32.970709+00:00*
 *Test results: 14 working, 1 failed, 1 skipped out of 16 total servers*
 
 | Server Name | Installation | Protocol | Status | Description |


src/agent_factory/utils/json_exporter.py
```diff
@@ -1,36 +1,37 @@
...
-
-        output_file.write_text(json.dumps(all_spans, indent=2))
+        output_file = self.output_dir / f"0x{format_trace_id(spans[0].context.trace_id)}.jsonl"
+
+        with output_file.open("a", encoding="utf-8") as f:
+            for span in spans:
+                # We don't need a2a server events in traces
+                if span.attributes.get(GenAI.OPERATION_NAME) in KEEP_SPANS_WITH_ANY_AGENT_OPERATION_NAME:
+                    try:
+                        agent_span = AgentSpan.from_otel(span)
```
HareeshBahuleyan, line 31: It would be great if we could collect all the agent_spans and create and [AgentTrace](https://github.com/mozilla-ai/any-agent/blob/e263ebf03958415f3b1de279ba206d4591470177/src/any_agent/tracing/agent_trace.py#L187) object at the end.

What I am not sure 
1. Is how to determine "the end"?
2. How to get the `final_output`?

daavoo (approved):

PR #266, opened 2025-08-14
refactor(artifact): adds litellm<1.75.0 to all target agents

### LINKED ISSUE #261: No module backoff found
As of today, running a target agent yields this error, which in turn is coming from this [known issue in LiteLLM](https://github.com/BerriAI/litellm/issues/13081):

`No module named 'backoff'. Run pip install 'litellm[proxy]'`

A quick fix seems to be pinning LiteLLM to < 1.75, which was known to work, both for the manufacturing agent and any target agents.

Long-term fix: pin all dependencies that are known to the manufacturing agent (e.g. match agent deps against pyproject.toml and use those versions in the target agent requirements.txt; bonus: replace that requirements.txt with better dependency handling).

ividal:
## 📝 What's changing

Only pins LiteLLM to fix the problem we're having in #261  and enable running the agents.

Full searching for the versions and pinning in #267 

> If this PR is related to an issue or closes one, please link it here.

Closes #261 

---

## 📚 How to test it

Steps to test the changes:

1. `make run`
2. `uv run chainlit run src/agent_factory/chainlit.py`
3. Create a new agent
4. Its requirements.txt should list litellm<1.75.0
5. Run the agent, no "No module backoff" error anymore

---

## ✅ Pre-merge checklist

- [X] Added some tests for any new functionality
- [X] Tested the changes in a working environment to ensure they work as expected
- [X] [Manually triggered](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow) workflows from this branch and ensured that test that involve agent generation using LLM API keys run successfully. Link to successful workflow run: [here](https://github.com/mozilla-ai/agent-factory/actions/runs/16978693650)

### DIFF
--- a/src/agent_factory/utils/artifact_validation.py
+++ b/src/agent_factory/utils/artifact_validation.py
@@ -17,14 +17,28 @@ def clean_python_code_with_autoflake(code: str) -> str:
         raise
 
 
+# TODO: pin all dependencies, if known, i.e. if present in the pyproject.toml
 def validate_dependencies(agent_factory_outputs: dict[str, str]) -> str:
     """Validate dependencies. In particular:
     - make sure that if uvx is used to install an MCP server, then
       uv appears in the final requirements.txt
+    - Any line starting with `litellm` (optionally with extras and any version
+      specifiers) is replaced with exactly `litellm<1.75.0`.
+    - If no `litellm` line exists and dependencies are non-empty, append
+      `litellm<1.75.0` as a new line.
     """
-    # make sure uv is in if uvx is in tools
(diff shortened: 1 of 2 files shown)


src/agent_factory/utils/artifact_validation.py
```diff
@@ -17,14 +17,28 @@ def clean_python_code_with_autoflake(code: str) -> str:
         raise
 
 
+# TODO: pin all dependencies, if known, i.e. if present in the pyproject.toml
```
ividal, line 20: Yes, this is dirty. Full version pinning in #267 .

PR #217, opened 2025-08-04
Post-mortem agent use-case

stefanfrench:
## 📝 What's changing

Add an agent that uses input data from Notion, Slack and Jira to create a post-mortem report on a Notion page about an internal incident.

Closes #216 

---

## 📚 How to test it

Steps to test the changes:

1. Follow the instructions in the [use-case guide](https://github.com/mozilla-ai/agent-factory/blob/216-add-example-agents-for-demos/docs/getting-started/post-mortem-agent.md)
2. Ask @Kostis-S-Z for Slack/Notion/Jira credentials 

---

## ✅ Pre-merge checklist

Please ensure the following items are checked **before merging** the PR (if not, please add a small explanation why).

- [ ] Added some tests for any new functionality
- [ ] Tested the changes in a working environment to ensure they work as expected
- [ ] [Manually triggered](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow) workflows from this branch and ensured that test that involve agent generation using LLM API keys run successfully. Link to successful workflow run: [insert-link-here]
  >Note: The can be done from [Actions tab in GitHub UI](https://github.com/mozilla-ai/agent-factory/actions/workflows/tests.yaml) -> Select the pinned `Tests` workflow from the left sidebar -> Select `your-branch-name` -> `Run workflow`

---

## 📸 Screenshots (if applicable)

Please add any relevant screenshots to show the changes (e.g. agent performance in comparison to `main` branch).

PR #288, opened 2025-08-20
Revamp integration tests with mocks

aittalam:
## 📝 What's changing

Provide a clear and concise description of the content changes you're proposing. List all the changes you are making to the content.

* Added new...
* Removed outdated information from...

> If this PR is related to an issue or closes one, please link it here.

Refs #...

Closes #...

---

## 📚 How to test it

Steps to test the changes:

1.
2.
3.

---

## ✅ Pre-merge checklist

---

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -159,3 +159,61 @@ jobs:
             make test-generated-artifacts PROMPT_ID="$prompt_id" || { echo "::error::Tests for PROMPT_ID $prompt_id failed!"; exit 1; }
             echo "::endgroup::"
           done
+
+
+  run-generated-artifacts-integration-tests:
+    if: github.event_name == 'workflow_dispatch'
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@v4
+
+      - name: Install the latest version of uv and set the python version to 3.13
+        uses: astral-sh/setup-uv@v6
+        with:
+          python-version: 3.13
+          activate-environment: true
+
+      - name: Cache mcpd binary
+        id: cache-mcpd
+        uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+        with:
+          key: mcpd-${{ vars.MCPD_VERSION }}-Linux-x86_64
+          path: /usr/local/bin/mcpd
+
+      - name: Install mcpd (if not cached)
+        if: steps.cache-mcpd.outputs.cache-hit != 'true'
+        run: |
+          version="${{ vars.MCPD_VERSION }}"
+          if [ -z "$version" ]; then
+            echo "MCPD_VERSION var is not set, or empty"
+            exit 1
+          fi
+          zip="mcpd_Linux_x86_64.tar.gz"
+          url="https://github.com/mozilla-ai/mcpd/releases/download/${version}/${zip}"
+          curl -fsSL "$url" -o "$zip" || { echo "download failed"; exit 1; }
+          tar -xzf "$zip" mcpd || { echo "extract failed"; exit 1; }
+          sudo mv mcpd /usr/local/bin/mcpd && sudo chmod +x /usr/local/bin/mcpd || { echo "Install failed"; return 1; }
+          rm -f "$zip"
+
+      - name: Run Generated Artifacts Integration tests
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          MCPD__ELEVENLABS_MCP__ELEVENLABS_API_KEY: "mock" # pragma: allowlist secret
+          MCPD__SLACK__SLACK_BOT_TOKEN: "mock" # pragma: allowlist secret
+          MCPD__SLACK__SLACK_TEAM_ID: "mock" # pragma: allowlist secret
+          MCPD__SQLITE__DB_PATH: "blueprints.db" # pragma: allowlist secret
+        run: |
+          PROMPT_IDS=(
+            url-to-podcast,
+            scoring-blueprints-submission
+          )
+          for prompt_id in "${PROMPT_IDS[@]}"; do
+            echo "::group::Running tests for PROMPT_ID: $prompt_id"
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yaml
```diff
@@ -159,3 +159,69 @@ jobs:
...
+            summarize-url-content
+            url-to-podcast
+            scoring-blueprints-submission
+          )
+          for prompt_id in "${PROMPT_IDS[@]}"; do
+            echo "::group::Running tests for PROMPT_ID: $prompt_id"
+            if [ -f "tests/artifacts/$prompt_id/secrets.prod.toml" ]; then
+              echo "[i] Starting mcpd as it is required for this agent"
+              mcpd daemon --runtime-file tests/artifacts/$prompt_id/secrets.prod.toml --config-file tests/artifacts/$prompt_id/.mcpd.toml &
+              sleep 2
```
aittalam, line 220: This is required for mcpd to pull up all the mcp servers it needs (without it the test starts before mcpd is up). I am not sure this will work for future use cases with potentially more servers, and this also can vary depending on networking... Perhaps we should already put a longer time?

tests/generated_artifacts/tool_mappings.py
```diff
@@ -7,20 +15,58 @@
...
-        "mock_function": mock_text_to_speech,
+        "type": "function",
+        "match_condition": {
+            "name": "extract_text_from_url",  # this string appears in the function name
+        },
+        "mock_function": mock_extract_text_from_url,
     },
     {
-        "prompt_id": "url-to-podcast",
+        "prompt_id": "url-to-podcast",  # mocks can be specific to a use-case (identified by its prompt_id)
```
HareeshBahuleyan, line 33: Having the comments for the first example in the list would suffice, no? WDYT?

tests/generated_artifacts/tool_mappings.py
```diff
@@ -42,58 +88,77 @@
 # ----------------------------------------------------------------------------
 
 
-def find_matching_validation(tool, tool_type_checker, prompt_id):
+def find_matching_validation(tool, prompt_id):
```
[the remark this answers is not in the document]
HareeshBahuleyan: +1 agree to remove the unused code.

HareeshBahuleyan (commented):
Thanks Davide.

1. Like you suggested, removing the parts of the mocks for MCPStdio which is not used would be great!

2. How do you see this for future use cases? Would the idea be that contributors of new use cases (such as #217 , #290) are also required to submit mocks? If yes, I wonder if should have that on the CI, to check if run-generated-artifacts-integration-tests exist for all prompt ids in use_cases.yaml (Feel free to ignore this if you think that we don't need such enforcement at the moment 😄 )

.github/workflows/tests.yaml
aittalam: uuuh that's a great idea, I'll check that out!

aittalam: Added a script in this commit: https://github.com/mozilla-ai/agent-factory/pull/288/commits/d32df95bfb1bc9304f21a03485b36b8d101cbe10

tests/generated_artifacts/tool_mappings.py
```diff
@@ -7,20 +15,58 @@
 TOOL_MOCKS = [
     {
         "prompt_id": "url-to-podcast",  # mocks can be specific to a use-case (identified by its prompt_id)
-        "type": "MCPStdio",  # Tool type: 'mcp' for MCPStdio tools, 'function' for regular functions
-        "match_condition": "mcp/elevenlabs",  # String to match in tool args (for MCP) or function name (for functions)
+        "type": "function",  # Tool type: 'mcp' for MCPStdio tools, 'function' for regular functions
```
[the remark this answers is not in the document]
aittalam: Fixed in https://github.com/mozilla-ai/agent-factory/pull/288/commits/35df4b16f3d234123a25c43e226c0b695fdb3d88

tests/generated_artifacts/tool_mappings.py
```diff
@@ -7,20 +15,58 @@
...
+        "match_condition": {
+            "name": "elevenlabs_mcp__text_to_speech",  # this string appears in the function name
```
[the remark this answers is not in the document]
aittalam: Good idea! Added better documentation in https://github.com/mozilla-ai/agent-factory/pull/288/commits/35df4b16f3d234123a25c43e226c0b695fdb3d88

aittalam:
>     1. Like you suggested, removing the parts of the mocks for MCPStdio which is not used would be great!

Thank you! Yeah the code is much simpler now, it's gonna be easier for ppl to tinker with it if needed 🙏 

>     2. How do you see this for future use cases? Would the idea be that contributors of new use cases (such as #217 , #290) are also required to submit mocks? If yes, I wonder if should have that on the CI, to check if run-generated-artifacts-integration-tests exist for all prompt ids in use_cases.yaml (Feel free to ignore this if you think that we don't need such enforcement at the moment 😄 )

I would love ppl to create their own integration tests for new agents! The reason is that these tests actually run the agents and make sure they behave in a predictable way, for instance verifying that some outputs are always there and they are the expected ones.

I would not enforce this _for now_, but as all of us know about these tests now we can raise a flag whenever we are reviewing a new agent so that who is adding that will create an ad-hoc test for it. The main reason for not enforcing is that the person who makes the agent might not be the same that prepares the test, at least until this becomes a shared practice that everyone knows how to complete. I am happy to pair with whoever wants to build some mocks of course ;-)

Another thing that I'd like to raise is that the StructuredOutput format can change whenever we rebuild an existing agent. So unless we want to constrain it (which I see as... complicated 😅) we should be aware that this might break these tests, as they rely on StructuredOutput to pick up the results and verify them. Happy to add some documentation for anyone who creates a new or rebuilds an existing agent!

tests/generated_artifacts/tool_mappings.py
```diff
@@ -1,99 +1,86 @@
-from generated_artifacts.tool_mocks import mock_extract_text_from_url, mock_text_to_speech
-from generated_artifacts.tool_validations import no_docker_mcp
+from any_agent.config import MCPStdio
```
HareeshBahuleyan, line 1:
```suggestion
```

HareeshBahuleyan (approved):
Tiny suggested change on potential unused import. 

LGTM, thanks for the changes.

Issue #256, opened 2025-08-13
Save manufacturing agent traces along with other agent artifacts


HareeshBahuleyan:
Currently they are saved in `traces/` folder.

We should be able to enable saving to the preferred location set up the user - s3/local/minio

Issue #192, opened 2025-07-28
Migrate tests to A2A - Part-4 Tests on Manufacturing Agent Trace


HareeshBahuleyan:
This is required to understand if changing instructions/model/framework - how it impacts token usage, #turns, etc.

Dependent on first completing: https://github.com/mozilla-ai/agent-factory/issues/256

PR #292, opened 2025-08-21
chore(deps): Updates to use latest `any-agent`.

### LINKED ISSUE #210: Factory imports any_agent[all], which is overkill
Factory could very well only install essential frameworks. Recommended would be tinyagent (faster feedback/fix loop) and openai for the manufacturing agent model.

### LINKED ISSUE #251: Replace default framework to be tinyagent instead of openai
Previously, an issue with `tinyagent` was that it ignored tools and generated the outputs without any tool calling.
This seems to have been fixed in any-agent v1.10 https://github.com/mozilla-ai/any-agent/releases/tag/1.1.0

Run some tests and based on results, switch over to `tinyagent` if positive

### LINKED ISSUE #268: Updates to llm-calling tools
A couple of updates to our llm-calling tools:

- remove litellm dependency in favor of any-llm
- change gpt-4o-mini to o3

### LINKED ISSUE #283: tinyagent does not return the Plan in multi-turn chat
When running multi-turn - for example via chainlit, `tinyagent` does not return the plan, but instead returns:

```
Your plan, steps, and listed tools are sufficient. Proceed with creating the agent.
```

This does not happen with `openai` which outputs:

```

```

I tried with a few different initial prompts and it seems to be the same. 

To replicate:
Terminal-1:
```
cd src/agent_factory && uv run . --host 0.0.0.0 --port 9095 --framework tinyagent
```
Terminal-2:
```
A2A_SERVER_PORT=9095 uv run chainlit run src/agent_factory/chainlit.py
```
On the chainlit UI on the browser, send a chat such as `Workflow to provide the expected weather tomorrow in a city provided by the user.`

daavoo:
### What's changing
- Drop extra deps `all`. Use `tinyagent` by default. Manually install extras if/when needed.

- The default behavior is to use manufacturing agent to `any-llm` for both manufacturing and target agent (`any-agent` now uses `any-llm` instead of `litellm` for `tinyagent`).

- Update `model` syntax (was missing `provider` prefix, which was ok for `litellm` but not for `any-llm`)

Fixes #210 - no need to install `any-agent[all]` for manufacturing agent
Fixes #251 - change default framework to `tinyagent`
Fixes #268 - Partially closes this issue, whats remaining has been moved to a new issue https://github.com/mozilla-ai/agent-factory/issues/317
Fixes #283 - tinyagent not returning the "Plan" in multi-turn chat

> Note: When reviewing, please ignore the files under artifacts/ as they had to be regenerated.

### Comparison of metrics
Check the comparison between the metrics with the 2 frameworks openai and tinyagent here:
https://github.com/mozilla-ai/agent-factory/pull/292#issuecomment-3227255366

### How to test
1. From the chainlit UI
OR
2. Using make:
    choose a $prompt-id from the three use cases we have (summarize-url-content, url-to-podcast, scoring-blueprints-submission) and run `make test-single-turn-generation-e2e PROMPT_ID=summarize-url-content`

Passing CI tests: https://github.com/mozilla-ai/agent-factory/actions/runs/17402568498/


docs/getting-started/usage.md
```diff
@@ -16,8 +16,8 @@ In addition to `host` and `port`, you can also pass the following arguments:
 
 -  `--chat` vs `--nochat`: `chat` mode enables multi-turn conversations, while `nochat` mode is for one-shot tasks (default:
    `chat`).
--  `--framework`: The Agent framework to use (default: `openai`).
--  `--model`: The model ID to use (default: `o3`).
+-  `--framework`: The Agent framework to use (default: `tinyagent`).
```
HareeshBahuleyan, line 19: When testing, while `tinyagent` works, the other frameworks do not work anymore (I assume this is due to the transition to any-llm?)

Dockerfile
```diff
@@ -10,11 +10,11 @@ FROM python:3.13-slim
 # Set the working directory in the container
 WORKDIR /app
 
-ENV FRAMEWORK=openai
+ENV FRAMEWORK=tinyagent
```
HareeshBahuleyan, line 13: Since we don't have experiments yet, I would suggest we wait for https://github.com/mozilla-ai/agent-factory/issues/192 and https://github.com/mozilla-ai/agent-factory/issues/256 to be completed before proceeding with migration to tinyagent.

docs/getting-started/usage.md
Kostis-S-Z: its because now the `pyproject.toml` only installs the bare minimum any-agent library (tinyagent + a2a support), and not any of the other frameworks. See [here](https://github.com/mozilla-ai/agent-factory/pull/292/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711).

I think this change actually might be too nuclear, and there is a high chance it will break the platform, as I think the expectation is that agent-factory supports all frameworks by default from a user perspective (and they dont have to manually other frameworks).

Either 
1. we do keep the core dependency only with tinyagent and add a group of `all-frameworks` (or something like that) that should be the default way of installing when used by the platform
2. keep the core dep with all frameworks supported

Even though I understand that all of these frameworks are adding a lot of weight, I think for now, to ensure stability, we stay with 2., and revisit this later.

Kostis-S-Z (changes requested):
Since we are tracking `uv.lock` in this repo, could you also push that?

docs/getting-started/usage.md
Kostis-S-Z: > For the manufacturing agent or for the manufactured agents?

I was thinking for both, but I re-read the [document](https://docs.google.com/document/d/199vjp9461-4OfPk96iwBOeeF95iFpFI5la4cAGXK2Vo/edit?tab=t.ahw7h4duagwp) and maybe you are right that its fine if tinyagent is supported by default for manufacturing? 
> 4.2 Enable the user to swap model provider and model for the target agent (from a limited subset: 2 model providers, 3 models total. Note:using any-llm/any-agent

In any-case I think its good if we loop in someone from the platform to make sure that this change doesnt break anything there.

> Not sure I follow this, how does it affect stability?

I meant that if there is an expectation from the platform that all frameworks are supported by default, changing this behaviour might break things. 

tl;dr lets sync with platform folk to ensure we can move ahead with this change.

Kostis-S-Z: > For the manufacturing agent or for the manufactured agents?

I was thinking for both, but I re-read the oct req document and maybe you are right that its fine if tinyagent is supported by default for manufacturing? 

In any-case I think its good if we loop in someone from the platform to make sure that this change doesnt break anything there.

> Not sure I follow this, how does it affect stability?

I meant that if there is an expectation from the platform that all frameworks are supported by default, changing this behaviour might break things. 

tl;dr lets sync with platform folk to ensure we can move ahead with this change.

Kostis-S-Z:
@daavoo 

Running this:
```
uv pip install 'any-agent[openai]'
uv run . --host 0.0.0.0 --port 8080 --nochat --framework=openai 
```

Returns this:
```
                             AgentRunError: litellm.BadRequestError: LLM Provider NOT                            
                             provided. Pass in the LLM provider you are trying to call. You                      
                             passed model=openai:o3                                                              
                              Pass model as E.g. For 'Huggingface' inference endpoints pass                      
                             in `completion(model='huggingface/starcoder',..)` Learn more:                       
                             https://docs.litellm.ai/docs/providers     
```

if I add `--model=openai/o3` it works. 

Am I understanding correctly that `any-llm` can parse both `<provider>:<model_id>` and `<provider>/<model_id>` (tested locally, both of these formats worked when starting the server with the CLI), but litellm can only parse `<provider>/<model_id>`?

Kostis-S-Z:
@daavoo also if we do want to maintain the docker image, then if we pass `FRAMEWORK=openai` during `make run`, then that should also update the venv inside the container to `pip install 'any-agent[openai]` :melting_face: which ofc is not great... so either in the image we keep `any-agent[all]` or remove the support of changing the framework through the `make run` args?

dpoulopoulos:
> > Am I understanding correctly that `any-llm` can parse both `<provider>:<model_id>` and `<provider>/<model_id>` (tested locally, both of these formats worked when starting the server with the CLI), but litellm can only parse `<provider>/<model_id>`?
> 
> That is correct, but slash (`/`) syntax will be deprecated.
> 
> This issue is that in `any-agent` we haven't made the changes yet to use `any-llm` for the `openai` framework.
> 
> I will make an update to this PR to use the slash syntax (until we fully replace `litellm` in `any-agent`)

@daavoo I think the issue was this: https://github.com/mozilla-ai/any-llm/pull/348

dpoulopoulos (commented):
@daavoo you should also change the `openai` framework references in `instructions.py`, right?

- https://github.com/mozilla-ai/agent-factory/blob/3bfd49551ac47821be007848cb6083c8e681ce09/src/agent_factory/instructions.py#L83
- https://github.com/mozilla-ai/agent-factory/blob/3bfd49551ac47821be007848cb6083c8e681ce09/src/agent_factory/instructions.py#L309

daavoo:
> @daavoo I think the issue was this: [mozilla-ai/any-llm#348](https://github.com/mozilla-ai/any-llm/pull/348)

No, in this case Kostis was switching the agent framework to `openai`. We still rely on `litellm` for that framework and the syntax used here by default `:` is not supported in `litellm`.

> @daavoo you should also change the `openai` framework references in `instructions.py`, right?
> 
>     * https://github.com/mozilla-ai/agent-factory/blob/3bfd49551ac47821be007848cb6083c8e681ce09/src/agent_factory/instructions.py#L83
> 
>     * https://github.com/mozilla-ai/agent-factory/blob/3bfd49551ac47821be007848cb6083c8e681ce09/src/agent_factory/instructions.py#L309

Since the original issue was about the "main" agent, I was not sure if we also want to update the "generated" agents, but I can include that change

src/agent_factory/start.sh
```diff
@@ -2,8 +2,8 @@
 set -eu
 
 # Set default values for environment variables if not set
-: "${FRAMEWORK:=openai}"
-: "${MODEL:=o3}"
+: "${FRAMEWORK:=tinyagent}"
```
HareeshBahuleyan, line 5: After pulling the latest changes from `main`, I run:
```
FRAMEWORK=tinyagent MODEL=openai:o3 make test-single-turn-generation-e2e PROMPT_ID=summarize-url-content
```

And get the metrics summary (but it does not contain the costs):
image

Dockerfile
HareeshBahuleyan: @daavoo Update: we now have the traces and a metrics table output at the end of the generation tests which makes it easier to compare models/frameworks https://github.com/mozilla-ai/agent-factory/pull/300#issuecomment-3223396789
^what I have there is run on any-agent v1.2.0, since I could not get the [costs row](https://github.com/mozilla-ai/agent-factory/pull/292#discussion_r2300857409) on this branch.

It would be great to have such a comparison before making the switch over to tinyagent.

src/agent_factory/start.sh
daavoo: by changing to / we still use any-llm for doing the LLM calls and just use litellm to get the pricing (which is basically a JSON lookup). the issue was that litellm was failing yo find the price using the : syntax.

HareeshBahuleyan:
Hey @daavoo 
I ran the [generation CI test](https://github.com/mozilla-ai/agent-factory/blob/8f0b1056761ae3ef9a7f8c5ad24fe8338119669c/.github/workflows/tests.yaml#L58) locally for url-to-podcast (10 runs, minimum 6 success expected) and get the following results:

Tinyagent (1 failure):
image

Openai (3 failures):
image

Although openai has lower token usage, n_turns and costs **on average per run**, I think we could switch to tinyagent for the manufacturing agent's default (given that we have more control of the changes in tinyagent framework).

Before proceeding, could we please do the following as well:

1. Update docs: https://mozilla-ai.github.io/agent-factory/getting-started/usage/ - which mentions openai as the default

2. Manually trigger generation [tests](https://github.com/mozilla-ai/agent-factory/actions/workflows/tests.yaml) from this branch and check if they pass

3. As @Kostis-S-Z noted [here](https://github.com/mozilla-ai/agent-factory/pull/292#discussion_r2298197920), how about adding to the generation tests - atleast one use case ALSO runs with openai framework + litellm (this would ensure some automated checks that this combination also works, atleast till litellm is fully dropped)

Sorry for the long comment!

HareeshBahuleyan:
Quick update: Fixed up failing tests, regenerated artifacts and they pass the CI: https://github.com/mozilla-ai/agent-factory/actions/runs/17402568498/

Remaining TODO: Add a generation test that runs on CI with openai framework + litellm

Dockerfile
```diff
@@ -35,6 +35,11 @@ ENV SETUPTOOLS_SCM_PRETEND_VERSION=${APP_VERSION}
 COPY --from=mcpd /usr/local/bin/mcpd /usr/local/bin/mcpd
 RUN chmod +x /usr/local/bin/mcpd
 
+# Install build-essential for CPython dependencies (fastuuid)
```
dpoulopoulos, line 38: Please, please can we document why this change here? I mean expanding the comment a bit. 🙏🏻

tests/artifacts/scoring-blueprints-submission/.mcpd.toml
```diff
@@ -1,8 +1,8 @@
 [[servers]]
   name = "slack"
   package = "npx::@modelcontextprotocol/server-slack@latest"
-  tools = ["slack_post_message", "slack_list_channels"]
-  required_env = ["SLACK_BOT_TOKEN", "SLACK_TEAM_ID"]
+  tools = ["slack_list_channels", "slack_post_message"]
```
dpoulopoulos, line 4: I don't get if there is a reason for reversing this list or it is just an aesthetic change.

pyproject.toml
```diff
@@ -5,12 +5,11 @@ version = "0.1.0"
 readme = "README.md"
 requires-python = ">=3.13"
 dependencies = [
-    "any-agent[all,a2a]==1.2.0",
+    "any-agent[a2a]==1.5.0",
     "autoflake==2.3.1",
     "boto3>=1.40.0",
     "python-dotenv>=1.1.1",
     "fire~=0.7.0",
-    "litellm<1.75.0",
```
dpoulopoulos, line 13: Installing `any-agent[a2a]==1.5.0` also installs:
- `any-llm-sdk>=0.15.0,<1`
- `litellm>=1.74.0`

My question is: why did we have `litellm` pinned to `<1.75.0`? Should we keep it pinned for this reason? Because the latest version of `any-agent` has it pinned to `litellm>=1.75.8` which could cause us issues.

HareeshBahuleyan: >Should we keep it pinned for this reason? 

Yes, this is also the reason that I did not bump to the more recent versions of any-agent.

Reason:
When testing with any-agent>0.15, it [requires](https://github.com/mozilla-ai/any-agent/blob/60076527b5cc7fa3e7d19373fc5988c6e6c292d9/pyproject.toml#L13) LiteLLM >1.75, but that results in an error for the target agent: See [Issue #13081](https://github.com/BerriAI/litellm/issues/13081) (=> we have to wait till LiteLLM fixes this issue)

I can add this reasoning as a comment here.

HareeshBahuleyan: Done in https://github.com/mozilla-ai/agent-factory/pull/292/commits/30b3ef49feaca354442e8d726eec3f57176ccc2a

Dockerfile
HareeshBahuleyan: Done in https://github.com/mozilla-ai/agent-factory/pull/292/commits/53f319a242961f270ce2aac6c834b5634753720b

dpoulopoulos (approved):

PR #297, opened 2025-08-22
Migrate tests to A2A - Part-4 Tests on Manufacturing Agent Trace

HareeshBahuleyan:
## 📝 What's changing

Since we now have the traces from the manufacturing agent after merging #256, this PR aims to bring back the dropped tests that run on agent factory trace.

In essence, we bring back:
1. the tests that existed under [test_generated_traces.py](https://github.com/mozilla-ai/agent-factory/blob/archive/main-20250721/tests/generated_artifacts/test_generated_traces.py) before sunsetting agent factory CLI, and
2. the tests that were dropped in commits https://github.com/mozilla-ai/agent-factory/pull/193/commits/e1a8b3dc37c66266ad717b01fff489cfd6c5b498

P.S. The artifacts for `scoring-blueprints-submission` usecase had to be regenerated: so you can skip reviewing the files used `artifacts/`

Closes #192 and #179 

---

## 📚 How to test it

Steps to test the changes:

Run the tests on already generated artifacts
```
make test-generated-artifacts PROMPT_ID=summarize-url-content
```

and

Run the tests as a part of single turn generation
```
make test-single-turn-generation-e2e PROMPT_ID=summarize-url-content UPDATE_ARTIFACTS=--update-artifacts
```

---

## ✅ Pre-merge checklist

Please ensure the following items are checked **before merging** the PR (if not, please add a small explanation why).

- [x] Added some tests for any new functionality
- [x] Tested the changes in a working environment to ensure they work as expected
- [x] [Manually triggered](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow) workflows from this branch and ensured that test that involve agent generation using LLM API keys run successfully. Link to successful workflow run: [[link-here](https://github.com/mozilla-ai/agent-factory/actions/runs/17160897895/job/48690294723)]

---

## 📸 Screenshots (if applicable)
The tests on traces would be able to provide info on num of steps and tokens used by the manufacturing agent. This can be used later when comparing to different versions of the instruction prompt or when changing agent framework/model.
image

We also now have in the CI logs the costs encountered for the n-runs of the generation tests:
image

PR #300, opened 2025-08-25
Metrics tracker with costs, n_turns and token counts displayed at the end of generation tests

HareeshBahuleyan:
## 📝 What's changing

Closes #297 

---

## 📚 How to test it

Steps to test the changes:
```
make test-single-turn-generation-e2e PROMPT_ID=summarize-url-content # or try with another prompt-id 
```
At the end of the test check for a table in the console output that looks like this:
image

---

## ✅ Pre-merge checklist

- [x] Added some tests for any new functionality
- [x] [Manually triggered](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow) workflows from this branch and ensured that test that involve agent generation using LLM API keys run successfully. Link to successful workflow run: [[link-here](https://github.com/mozilla-ai/agent-factory/actions/runs/17208112381)]

---


aittalam (approved):
Thanks Hareesh, this is a very useful upgrade!

The code looks good to me, I ran it and the cost code was working nicely. I had an issue with the time the agents took to complete so the test failed, but I think that's something I should report in your other PR. This one is good to go!

HareeshBahuleyan:
Comparison across frameworks on `url-to-podcast`:

### Experiment setup:
```
max_attempts: 10
min_successes: 6
```
* any-agent 1.2.0
* all frameworks were using LiteLLM 
* all frameworks were using o3 as the model

---

### Results:

| Framework | OpenAI | Tinyagent | Langchain |
|:--------------:|:------:|:---------:|:---------:|
| Failed Runs    | 0 | 2 | 1 |
| Image          | openai | tinyagent | langchain |

Costs: tinyagent > openai > langchain
Time: openai > tinyagent > langchain

Failed runs in tinyagent - sometimes the artifacts did not satisfy the post-generation checks, e.g. MCP server used, requirements.txt valid, etc.

❗ Caveat: This table alone is not enough since it does not consider how well the target agent performs to accomplish the task!

aittalam:
> > Q: I assumed the number of runs in the table caption is successful_runs + failed_runs (e.g. 6+0 for oai, 6+2 for tinyagent...). Why is it 6 for langchain if we need 6 successes and there is one failed run? I'd assume it'd be 7.
> 
> Great observation 😄 - the reason has to do with where the metrics get added:
> 
> https://github.com/mozilla-ai/agent-factory/blob/91855db2410c8a9285b77dc78e320b5fdc6e6680/tests/generation/test_single_turn_generation.py#L188
> 
> It happens AFTER generation itself succeeds, `_assert_execution_time_within_limit`, etc. but BEFORE `validate_generated_artifacts` which runs the tests on the generated artifacts (agent_trace, agent_code). So some failure could happen after metrics_tracker is updated.
> 
> I could move the `metrics_tracker` to the very end. The only pre-requisite is that it need to be **after** the check that there is a valid agent_trace that has been saved.
> 
> By moving to the end, we lose appending run metrics (e.g. costs encountered) if the `validate_generated_artifacts` assertion fails, which I think could be useful (to get an understanding of how much $ was spend overall on success and failed runs that atleast had a trace). But I also see your point that the inconsistency could cause confusion.

I see! Well, I agree with you that it'd be way better to have more metrics, so let's keep it like this for now 👍

PR #301, opened 2025-08-25
feat: Automate requirements.txt generation from agent code

### LINKED ISSUE #121: Build `requirements.txt` by parsing the generated `agent.py`
We currently ask the agent to build the requirements.txt, which implies potential hallucinations like missing or hallucinated imports (which we have to test #112 ).

Instead of asking the agent to generate the `requirements.txt` file, we can instead parse the generated `agent.py` using https://docs.python.org/3/library/ast.html#ast.parse and manually build the requirements.txt.

I believe this would make the process more reliable but also help the agent by reducing the instructions size and requiring it to do one less thing.

dpoulopoulos:
## 📝 What's changing
This PR refactors the agent generation process to improve the reliability of dependency management. Previously, the manufacturing agent was prompted to generate the generated agent's dependencie itself, which was prone to hallucinations and could lead to missing or incorrect dependencies.

This new approach removes this responsibility from the manufacturing agent and instead generates extracts the dependencies by parsing the generated agent's code (`agent.py`).

## Key Changes

1. Remove Dependency Generation from Agent:
   * The `dependencies` field has been removed from the agent's output schema (`AgentFactoryOutputs`).
   * All instructions, prompts, and examples asking the agent to create a list of dependencies have been removed.
2. Introduces Automatic Generation Logic:
   * Parse the generated `agent.py` using Python's `ast` module to identify all import statements.
   * Filter out standard library modules.
   * Use `importlib.metadata` to map the discovered import names to their correct PyPI package names (e.g., `bs4` -> `beautifulsoup4`).
   * Pin the `any-agent` version in the final `requirements.txt` to match the version used in the manufacturing agent.

TODO: Remove the `LiteLLM` dependency when #268 is merged.

Refs #112
Closes #121

---

## 📚 How to test it

Generate an agent using any of the existing use cases. A valid `requirements.txt` file should be part of the generated agent's package.

---

## ✅ Pre-merge checklist


src/agent_factory/utils/io_utils.py
```diff
@@ -38,6 +42,74 @@ def parse_cli_args_to_params_json(cli_args_str: str) -> str:
...
+
+    if "tools" in third_party_imports:
+        third_party_imports.remove("tools")
+
+    logger.info("Mapping import names to their corresponding package names...")
+    # packages_distributions() returns a mapping of top-level modules to the
+    # distribution (installable) package names that provide them.
+    # e.g., {'bs4': ['beautifulsoup4'], 'sklearn': ['scikit-learn']}
+    try:
+        dist_map = importlib.metadata.packages_distributions()
```
HareeshBahuleyan, line 94: Clarification Q: this will return a mapping for _only_ packages those are installed in this python environment, right?

src/agent_factory/utils/io_utils.py
```diff
@@ -38,6 +42,74 @@ def parse_cli_args_to_params_json(cli_args_str: str) -> str:
...
+def extract_requirements_from_string(python_code: str) -> list[str]:
+    """Extract dependencies from a string of Python code.
+
+    Analyze a string of Python code, filter out standard library modules, and extract the third-party imports.
+
+    Args:
+        python_code (str): The Python code to analyze.
+    """
+    stdlib = set(sys.stdlib_module_names)
+    all_imports = get_imports_from_string(python_code)
```
HareeshBahuleyan, line 83: Do we also need to run this on all python `tools` selected by the manufacturing agent, to be able to capture their imports + dependencies?

HareeshBahuleyan (commented):
Thanks @dpoulopoulos!
Couple of clarification questions from my side.

src/agent_factory/utils/io_utils.py
dpoulopoulos: Good idea, let me check.

src/agent_factory/utils/io_utils.py
```diff
@@ -46,20 +118,25 @@ def prepare_agent_artifacts(agent_factory_outputs: dict[str, str]) -> dict[str,
...
+    dependencies = []
     # Identify and read tool files
     for tool_file in TOOLS_DIR.iterdir():
         if tool_file.is_file() and (tool_file.stem in agent_code or tool_file.name == "__init__.py"):
-            artifacts_to_save[f"tools/{tool_file.name}"] = tool_file.read_text(encoding="utf-8")
+            tool_code = tool_file.read_text(encoding="utf-8")
+            dependencies.extend(extract_requirements_from_string(tool_code))
+            artifacts_to_save[f"tools/{tool_file.name}"] = tool_code
 
     cli_args_str = agent_factory_outputs.get("cli_args", "")
```
HareeshBahuleyan, line 133: Sorry, nitpick: Could we move the `cli_args_str` lines below, so that we can have all the dependency extraction code together.

src/agent_factory/utils/io_utils.py
```diff
@@ -46,20 +118,25 @@ def prepare_agent_artifacts(agent_factory_outputs: dict[str, str]) -> dict[str,
...
     artifacts_to_save["agent_parameters.json"] = parse_cli_args_to_params_json(cli_args_str)
 
+    dependencies.extend(extract_requirements_from_string(clean_agent_code))
```
HareeshBahuleyan, line 136: Should dependencies be a set instead of a list? Just to avoid duplicates if if the tool_code and clean_agent_code have common dependencies?

HareeshBahuleyan (approved):

Issue #176, opened 2025-07-18
Documenting running of tests locally during dev


HareeshBahuleyan:
Based on https://github.com/mozilla-ai/agent-factory/pull/165#discussion_r2215497995

When tests converge (after migrating tests to agent-factory API), it would be good to have a dev-oriented README to share info on how to run tests manually.  

Even better could be have a Makefile with different test setups as we had for Lumigator.

PR #313, opened 2025-08-31
Local tests documentation

aittalam:
## 📝 What's changing

* Added new README.md file inside `tests/` that explains how tests work and how they can be run locally or in CI.
* Updated Makefile to automatically run mcpd and check its services are up (as we already did in CI).
* Updated CI script to simplify + use recipe from Makefile.

Refs #176 

---

## 📚 How to test it

Go through the readme and follow instructions.

---

## ✅ Pre-merge checklist

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,24 @@ test-generated-artifacts-integration: check-prompt-id-present ## Run artifact in
 	@uv run --group tests pytest tests/generated_artifacts/ -m artifact_integration --prompt-id=$(PROMPT_ID) -v
 	@echo "Artifact integration tests completed successfully!"
 
+# pragma: allowlist secret
+test-generated-artifacts-integration-e2e: MOCK_TOKEN = mock
+test-generated-artifacts-integration-e2e: check-prompt-id-present
+	@if [ -f "tests/artifacts/$(PROMPT_ID)/secrets.prod.toml" ]; then \
+		echo "Starting mcpd as it is required for this agent"; \
+		export MCPD__ELEVENLABS_MCP__ELEVENLABS_API_KEY="$(MOCK_TOKEN)"; \
+		export MCPD__SLACK__SLACK_BOT_TOKEN="$(MOCK_TOKEN)"; \
+		export MCPD__SLACK__SLACK_TEAM_ID="$(MOCK_TOKEN)"; \
+		export MCPD__SQLITE__DB_PATH="blueprints.db"; \
+		mcpd daemon --runtime-file tests/artifacts/$(PROMPT_ID)/secrets.prod.toml --config-file tests/artifacts/$(PROMPT_ID)/.mcpd.toml & \
+		MCPD_PID=$$!; \
+		trap "kill $$MCPD_PID 2>/dev/null" EXIT; \
+		sh scripts/wait_for_mcpd_servers.sh && $(MAKE) test-generated-artifacts-integration PROMPT_ID="$(PROMPT_ID)" || { echo "::error::Tests for PROMPT_ID $(PROMPT_ID) failed!"; exit 1; }; \
+		kill $$MCPD_PID 2>/dev/null; \
+	else \
+		$(MAKE) test-generated-artifacts-integration PROMPT_ID="$(PROMPT_ID)" || { echo "::error::Tests for PROMPT_ID $(PROMPT_ID) failed!"; exit 1; }; \
+	fi
+
 # ====================================================================================
 # MCP Testing and Documentation
 # ====================================================================================
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,173 @@
+# Testing Guide
+
+We have a variety of tests implemented as a support to Agent Factory development.
+This guide describes them in detail and explains how to run them locally.
+The following sections are:
+
+- **Directory Structure**, explaining what is available and where.
+- **Running Tests**, describing how to run each different type of test.
+- **Tests in CI**, showing how tests are run on GitHub CI.
+
+## Directory Structure
+
+The `tests` folder is organized in the following sub-directories:
+
+- **`unit`**: unit tests for core functionality including e.g. a2a client/server,
+agent generation and validation, io and storage.
+- **`tools`**: tests for MCP server discovery and our built-in MCP tools.
+- **`generation`**: tests for end-to-end single-turn agent generation using real prompts. We have a few use cases detailed in the `use_cases.yaml` file, and the purpose of these tests is to verify that the process of building these agents completes with no errors.
+- **`generated_artifacts`**: these tests validate the agents generated in the previous step, making sure they are consistent with their input prompts and that they work as expected. The tests both verify the generated agents and their generation traces statically, and dynamically execute the agents while mocking some external APIs and paid services.
+- **`generated_agent_evaluation`**: our generated agents can also come with LLM-as-judge evaluation. The purpose of these tests is to validate the evaluation tool itself.
+- **`artifacts`**: these are not tests, but the actual agents generated in `generation` and tested in `generated_artifacts`.
+- **`utils`** - Shared testing utilities and helpers. At the present time, the package contains just a single `run_until_success_threshold_async` function, which allows us to run a test multiple times and consider it successful only if it completes at least n runs out of m attempts.
+
+While you can run the above directly with `pytest`, different tests might require different setups so we have added some shortcuts in our `Makefile` to run them in a single command.
+
+> NOTE: there is another set of tests pertaining the MCP servers we are curating in the `docs/mcp-servers.json` file: these do not validate our own code, but the MCP servers themselves. You can find the code in `docs/scripts/test_mcp_servers.py` and you can run the tests with `make test-mcps`.
+
+
+## Running Tests
+
+### Unit Tests
+
+These tests are the simplest ones, and you can run them with:
+
+```bash
+make test-unit
+```
+
+The `test-unit` target runs the following:
+
+```
+uv run --group tests pytest -v tests/unit/
+uv run --group tests pytest -v tests/tools/
+```
+
+which will first make sure all the libraries in the `tests` dependency group are present in the environment, the run `pytest` on the two directories `unit/` and `tools/`.
+
+### Generation Tests
+
+These tests replicate the full agent generation workflow with specific prompts. To do this, they need the A2A server to be up and running.
+
+The `pytest` command that is run by the `test-single-turn-generation` Makefile target is:
+
+```
+uv run --group tests pytest -xvs tests/generation/test_single_turn_generation.py --prompt-id=$(PROMPT_ID) $(UPDATE_ARTIFACTS)
+```
+
+This accepts two input parameters:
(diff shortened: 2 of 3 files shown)


tests/README.md
```diff
@@ -0,0 +1,173 @@
...
+- **`generated_agent_evaluation`**: our generated agents can also come with LLM-as-judge evaluation. The purpose of these tests is to validate the evaluation tool itself.
+- **`artifacts`**: these are not tests, but the actual agents generated in `generation` and tested in `generated_artifacts`.
+- **`utils`** - Shared testing utilities and helpers. At the present time, the package contains just a single `run_until_success_threshold_async` function, which allows us to run a test multiple times and consider it successful only if it completes at least n runs out of m attempts.
+
+While you can run the above directly with `pytest`, different tests might require different setups so we have added some shortcuts in our `Makefile` to run them in a single command.
+
+> NOTE: there is another set of tests pertaining the MCP servers we are curating in the `docs/mcp-servers.json` file: these do not validate our own code, but the MCP servers themselves. You can find the code in `docs/scripts/test_mcp_servers.py` and you can run the tests with `make test-mcps`.
```
HareeshBahuleyan, line 26: Should we directly link to the docs page for testing mcp servers? https://mozilla-ai.github.io/agent-factory/user-guide/mcp-servers/#testing-and-maintenance

tests/README.md
```diff
@@ -0,0 +1,173 @@
...
+These tests are the simplest ones, and you can run them with:
+
+```bash
+make test-unit
+```
+
+The `test-unit` target runs the following:
+
+```
+uv run --group tests pytest -v tests/unit/
```
HareeshBahuleyan, line 42: Regarding the `uv run` commands under each section, since they can be found via the makefile - do you think we could skip explicitly mentioning them here?

That way it also requires less maintenance - otherwise a change in makefile requires an update here too.

tests/README.md
```diff
@@ -0,0 +1,173 @@
...
+- `$(UPDATE_ARTIFACTS)` which is just a flag (`--update-artifacts`), to be added if you want the artifacts generated during this test to replace the ones which are already stored in the `artifacts/` directory.
+
+Note that this target assumes you are already running the A2A server and will break if you are not! The following two targets are more suitable as they first check whether the server is running and will not even start the tests otherwise:
+
+```bash
+make test-single-turn-generation-local PROMPT_ID=<prompt-id>
+make test-single-turn-generation-e2e PROMPT_ID=<prompt-id>
+```
+
+The first target just verifies the server is up, and is most convenient in the local setting where you usually alreay have the server running in background. The second target is mostly used in CI and will always start and stop a server for you.
```
HareeshBahuleyan, line 69:
```suggestion
The first target just verifies the server is up, and is most convenient in the local setting where you usually already have the server running in background. The second target is mostly used in CI and will always start and stop a server for you.
```

tests/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+The `pytest` command that is run by the `test-single-turn-generation` Makefile target is:
+
+```
+uv run --group tests pytest -xvs tests/generation/test_single_turn_generation.py --prompt-id=$(PROMPT_ID) $(UPDATE_ARTIFACTS)
+```
+
+This accepts two input parameters:
+- `$(PROMPT_ID)` which is one of the use-case names found in `generation/use_cases.yaml` (currently `summarize-url-content`, `url-to-podcast`, and `scoring-blueprints-submission`).
+- `$(UPDATE_ARTIFACTS)` which is just a flag (`--update-artifacts`), to be added if you want the artifacts generated during this test to replace the ones which are already stored in the `artifacts/` directory.
```
HareeshBahuleyan, line 60: Couple of suggestions on this:
A recommendation about when to update the artifacts and push them (e.g. expected change in agent.py code template, new artifact being generated, etc.)

Also a note about how updating them could affect the Artifact Tests (integration) below.

tests/README.md
```diff
@@ -0,0 +1,173 @@
...
+```
+uv run --group tests pytest -v tests/unit/
+uv run --group tests pytest -v tests/tools/
+```
+
+which will first make sure all the libraries in the `tests` dependency group are present in the environment, the run `pytest` on the two directories `unit/` and `tools/`.
+
+### Generation Tests
+
+These tests replicate the full agent generation workflow with specific prompts. To do this, they need the A2A server to be up and running.
```
HareeshBahuleyan, line 50: Should we mention these require API keys?

tests/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+```
+uv run --group tests pytest tests/generated_artifacts/ -m artifact_validation --prompt-id=$(PROMPT_ID) -v
+```
+
+Note the `-m artifact_validation` parameter: this is a *marker* that is used to group tests under the same namespace so you can follow them more easily in CI. Our namespaces are defined in `/pytest.ini`.
+
+### Artifact Tests (integration)
+
+These tests run the agents that have been created by *single-turn-generation* and verify they behave as expected.
```
HareeshBahuleyan, line 99:
```suggestion
These tests run the target agents that have been created by *single-turn-generation* and verify they behave as expected.
```

tests/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+In the more likely scenario where you need mcpd, you can run:
+
+```bash
+make test-generated-artifacts-integration-e2e PROMPT_ID=<prompt-id>
+```
+
+#### Q: What should I do if I add a new / change an existing agent artifact?
+
+Artifact tests -and in particular those with mocks as they run their actual code- are tightly dependent on
```
HareeshBahuleyan, line 125:
```suggestion
Artifact tests, and in particular those with mocks as they run their actual code- are tightly dependent on
```

tests/README.md
HareeshBahuleyan: Ahh, about the first point `when to update the artifacts `, I see you mentioned it below. Maybe we can move it up here? WDYT?

tests/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Tests in CI
+
+Almost all of our tests also run in CI. The CI pipeline (`.github/workflows/tests.yaml`) runs the following:
+
+- **Always Run**
+    - **Unit Tests** (`run-unit-tests`): fast unit tests on every push/PR
+    - **Generated Artifacts Tests** (`run-generated-artifacts-tests`) - static artifact validation tests
+
+- **Main Branch & Manual Only**
```
HareeshBahuleyan, line 168:
```suggestion
- **Main Branch & Manual Trigger Only**
```

HareeshBahuleyan: Suggest to add the reasoning behind why only on manual trigger -- because they require API keys and incur costs.

HareeshBahuleyan (commented):
Thanks Davide! This is super helpful. I liked the FAQ-style debugging at the end ❤️ 

Added a few minor comments and re-wording suggestions. I did a bit of refactor of eval tests in https://github.com/mozilla-ai/agent-factory/pull/311 and switched the model to o3 there. This might mean small modifications required for the current docs in the PR. Sorry for that!

tests/README.md
aittalam: Good idea, I think talking about this here makes sense and I can either move text or refer to it - even better to point to these concepts in multiple places in case people do not read the whole doc from beginning to the end

tests/README.md
aittalam: Yes! I will 🙏

tests/README.md
aittalam: I took the liberty of updating your own suggestion to completely remove dashes :-) I swear I wrote it myself btw 🤣

tests/README.md
aittalam: Good point, I will!

tests/README.md
```diff
@@ -0,0 +1,173 @@
...
+- **Running Tests**, describing how to run each different type of test.
+- **Tests in CI**, showing how tests are run on GitHub CI.
+
+## Directory Structure
+
+The `tests` folder is organized in the following sub-directories:
+
+- **`unit`**: unit tests for core functionality including e.g. a2a client/server,
+agent generation and validation, io and storage.
+- **`tools`**: tests for MCP server discovery and our built-in MCP tools.
```
dpoulopoulos, line 17: This is not that important, but why do we keep the tests for `tools` in a different directory? Aren't they just unit tests?

tests/README.md
```diff
@@ -0,0 +1,173 @@
...
+uv run --group tests pytest tests/generated_artifacts/ -m artifact_validation --prompt-id=$(PROMPT_ID) -v
+```
+
+Note the `-m artifact_validation` parameter: this is a *marker* that is used to group tests under the same namespace so you can follow them more easily in CI. Our namespaces are defined in `/pytest.ini`.
+
+### Artifact Tests (integration)
+
+These tests run the target agents that have been created by *single-turn-generation* and verify they behave as expected.
+Some of the MCP tools are mocked so we don't depend on external applications or paid APIs for our tests.
+As most MCP tools are managed by `mcpd`, its deamon needs to be started beforehand and passed the generated
```
dpoulopoulos, line 101: Instead of "passed the generated `.mcpd.toml` and `secrets.prod.toml` files" here, is it better to just say:

"Set the variables in the `.env` file that is generated by mcpd` and run the daemon"?

Makefile
```diff
@@ -144,6 +144,24 @@ test-generated-artifacts-integration: check-prompt-id-present ## Run artifact in
 	@uv run --group tests pytest tests/generated_artifacts/ -m artifact_integration --prompt-id=$(PROMPT_ID) -v
 	@echo "Artifact integration tests completed successfully!"
 
+# pragma: allowlist secret
+test-generated-artifacts-integration-e2e: MOCK_TOKEN = mock
```
dpoulopoulos, line 148: I would really love a comment here documenting how this mock works! 🙏🏻

tests/README.md
aittalam: Added in https://github.com/mozilla-ai/agent-factory/pull/313/commits/a2515ea60edc162aa42db22439c088c500a06a43

tests/README.md
aittalam: Fixed in https://github.com/mozilla-ai/agent-factory/pull/313/commits/b56a145cc5ffbab6d082cca97824e9869214a311

tests/README.md
aittalam: Fixed [here](https://github.com/mozilla-ai/agent-factory/pull/313/commits/cd6f87905424eab41d63761a11e88f31fd99b20e)

tests/README.md
aittalam: Done: https://github.com/mozilla-ai/agent-factory/pull/313/commits/01e41f518aefb8b4c715b9c8af9693085a7f7634

Makefile
aittalam: Added an explanation in [this commit](https://github.com/mozilla-ai/agent-factory/pull/313/commits/a8d5e6433ad2e290b4bacae0b2e3b761e88fedc8), lmk how that sounds to you!

tests/README.md
dpoulopoulos: Thanks for the thorough explanation!

